PR #48, opened 2024-10-18
fix: disabled elements don't follow the same style

### LINKED ISSUE #45: The disabled elements don't follow the same style
mpv_nFbES6YBTb

i.e. when audio track is unavailable, the volume icon and slider appears less visible/dark gray.

But for the audio and subtitle icons, the button becomes a little less bright but not nearly to the same extent as the volume ones, if that makes sense.

Not sure if they should be adjusted equally to match one or the other, or would a darker appearance look kind of weird/hard to see.

Xurdejl:
fixes #45


Samillion:
Well done in finding that, thank you very much.

PR #50, opened 2024-10-18
feat: dynamic time format based on video duration

Xurdejl:
I noticed the time display was always showing hours even for short videos (like 00:25:59 for a 25-minute video), which felt a bit cluttered and always bothered me. Most other players like YouTube only show hours when needed, so I thought we could do the same.

### What's changed?
- Short videos: 03:24 instead of 00:03:24
- Long videos: still 01:23:45 (no change)
- Everything else (milliseconds toggle, remaining time) works exactly the same

Let me know what you think!

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -2059,41 +2059,64 @@ local function osc_init()
         end
     end
 
-    -- tc_left (current pos)
-    ne = new_element("tc_left", "button")
-    ne.content = function ()
+    -- Helper function to format time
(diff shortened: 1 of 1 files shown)


Anthonyy232:
Would be a dope change

Samillion:
Absolutely love it. Thank you very much.

Samillion:
I need to update the preview images/video to showcase this change!

PR #61, opened 2024-10-20
Feat: Add locales as an optional separate file with english fallback

Xurdejl:
This PR adds support for an optional modernz-locale.lua file which contains all the locales, so users can now easily switch between languages using the language option in modernz.conf. 

### Changes

- Separate locales file `modernz-locale.lua`
- Defaults to English when not using the locales file so it works just like before
- Missing strings will fall back to English


Samillion:
Absolutely beautiful, this makes locale integration seamless. Thank you very much.

PR #76, opened 2024-10-22
Allow osc to hide properly with bottomhover off

### LINKED ISSUE #75: Keep osc visible while hovered during playback
During playback the osc always hides according to `hidetimeout=` even if hovered by cursor. In edge case scenario, if someone set `hidetimeout=0` the osc is almost unusable while playing.

Paused playback is not affected (i.e. the osc stays while hovered). If you observe stock OSC and youtube, their osc also stays visible while playing and hovered.

Request: change default behavior or add option to toggle behavior to keep osc visible while elements are hovered

Keith94:
Fixes #75


Samillion:
Thank you very much. I adjusted it to remove the parentheses.

Keith94:
Lol thanks! I was just thinking it looked weird

Issue #85, opened 2024-10-23
Touchpad triggers multiple inputs


Keith94:
The bug was reported here https://github.com/zydezu/ModernX/issues/52 (closed) but still persists in modernx/modernz.

- Only the skip and jump buttons seem to exhibit the issue (`showskip=yes` or `showjump=yes`).
- Only happens when using a laptop touchpad and tapping the touchpad.
- The tactile physical clicks don't cause the issue.
- The default OSC doesn't cause the issue either.

the log captures it but doesn't capture my inputs

[log.txt](https://github.com/user-attachments/files/17484698/log.txt)

[2.308][v][cplayer] Set property: chapter -> 1
[...]
[2.483][v][cplayer] Set property: chapter -> 1

Samillion:
So it triggers two actions per tap, right?

I'll investigate this and get back to you.

Keith94:
Yes, exactly that. mpv handles it flawlessly so maybe they have something in their codebase, idk.

Samillion:
I wonder if it's counting the touch down and up as two downs. Could you do a favor, find the line:
```lua
ne.eventresponder["mbtn_left_down"] = command_callback(user_opts.ch_next_mbtn_left_command)
```

change it to:
```lua
ne.eventresponder["mbtn_left_up"] = command_callback(user_opts.ch_next_mbtn_left_command)
```

Then try to skip forward chapter, still triggers two actions?

Keith94:
Hmm I can't test it properly because it breaks the script (lua error)

![image](https://github.com/user-attachments/assets/38ce8e5d-0264-4566-896a-7b1850145938)

Samillion:
Oops lol. I'll create a branch and link ya, sorry about that.

Keith94:
Aha that's fine. Not sure why it just broke or if I did something wrong.

Samillion:
Could you try the skip/jump buttons with this please:
https://github.com/Samillion/ModernZ/blob/dev_touchpad_up/modernz.lua

Keith94:
With a regular mouse it will appear normal, but only touchpad input crashes it.

Samillion:
😱  oh wow, ok, could you run the following please:

```
mpv --input-test --idle --force-window
```

In that mode, when you tap, what keybind is it using? and with a  single tap, what does it show/trigger?

Samillion:
Sorry, forgot to say to test it with the normal modernz, not the branch I linked.

[...]

PR #89, opened 2024-10-23
Fix skip and jump buttons activating twice

Keith94:
Fixes https://github.com/Samillion/ModernZ/issues/85

When I remove the line of code, the bug is fixed.

Disclaimer: I don't know what it precisely does or if it breaks anything.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -812,7 +812,6 @@ local function render_elements(master_ass)
 
                     element.eventresponder[state.active_event_source.."_down"](element)
                 end
-                state.mouse_down_counter = state.mouse_down_counter + 1
             end
         end
         


Samillion:
The code is correct, but something is triggering it twice. Could you do me a favor, could you comment out this line and try:

https://github.com/Samillion/ModernZ/blob/5b8a2b6b24746f1a986995606d14e9b87395fffb/modernz.lua#L1680

So it would be:
```lua
    --update_tracklist()
```

Does the double tap still persist?

Keith94:
>Does the double tap still persist?
Yeah, it does.

When you hold mouse down on the jump button, does the video keep jumping for as long as its held? That's what's causing my issue. Chapter buttons have the same behavior. But with stock osc it doesn't occur

Samillion:
Yeah, that's what `ne.softrepeat = true` does, which is set for jump/chapter buttons so that if you hold it down, it keeps seeking.

Samillion:
I have two options here, either set a debounce limit with a short timer (ie: 0.3), to make sure that only one clicks occur, but this might upset the people that hold down to seek.

The other option is turn softrepeat into a user option, so that you can enable it or not as you like.

Before doing that. Could you set softrepeat to false and try, please.

Keith94:
Sure. When soft repeat was off, there were no double inputs when tapping with trackpad.

Keith94:
For now I'll close this as invalid

PR #103, opened 2024-10-26
Feat: add windowcontrols_title user option

Xurdejl:
This is a user option from the vanilla mpv osc.


Samillion:
Thank you very much. Merging.

I wonder why they decided to add this option, I can't think of a scenario.

Issue #104, opened 2024-10-26
Seekbar tooltip always uses dynamic time


Keith94:
`time_format=fixed` properly affects the UI, but not the seekbar; the HH:  is always omitted for durations under an hour.

![image](https://github.com/user-attachments/assets/baff22a3-5353-4b22-8a1b-19afa35abcd7)

PR #105, opened 2024-10-26
fix: sync seekbar time format with time_format option

FreeTheTech101:
As requested in #104, this makes it so the time shown by the seekbar is affected by the `time_format` property.

Preview:
image


Samillion:
Thank you very much

PR #111, opened 2024-10-26
refactor: don't show osc on start when osc_on_seek=yes

Samillion:


Keith94:
I can't get it to work :( any ideas?

Samillion:
Are you sure? The behavior now is that it doesn't show OSC on start/next file, only if the OSC already showing it will not explicitly hide it, it'll just rely on hidetimeout then.

https://github.com/user-attachments/assets/a95a5772-51dc-49a1-b4b8-08879b01250f

Keith94:
Oh sorry you're right, it's my mistake. Works beautifully now!

Keith94:
Okay I just realized why it wasn't working for me.

If you have `save-position-on-quit` enabled, load and quit a file, the load the same file, the osc shows on startup because the player is seeking to the last position it was on before quitting.

Can ModernZ detect that?

Samillion:
I'll have to think about that one. Since technically with that scenario, it literally seeks back to position. So the event "seeking" is triggered.

PR #122, opened 2024-10-29
add hovereffect `size`

Commander07:
This PR adds an alternate style for hovered buttons.

## Changes
- `hovereffect` is no longer a boolean but a string corresponding to a hover effect (`glow`, `size`). Deafault is `glow`
- `hovereffect` is now disabled by leaving it blank or setting it to an unsupported value.
- Added the `size` hover effect which scales hovered buttons
- Added `hover_button_size` which is the scale applied to hovered buttons when `hovereffect` is set to `size`. Default is 115%

## Preview
https://github.com/user-attachments/assets/17f71e17-ac33-471e-9851-01c118809eca


Samillion:
That is absolutely beautiful. I've changed it so that hovereffect size is default. Thank you very much.

PR #123, opened 2024-10-29
fix: osc scaling behavior

FreeTheTech101:
This PR updates the OSC scale calculation code to match upstreams, ensuring that the `osd-scale-by-window` property is respected. This also requires changing `vidscale`'s default from `true` to `auto`. However, since `osd-scale-by-window` defaults to `true`, there are no changes to users with the default config.

**Changes**
---
`vidscale` now defaults to `"auto"`, which respects the `osd-scale-by-window` property (which by default is `true`)

**Preview**
---
| auto                                                                                               | yes                                                                                               | no                                                                                               |
|----------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------|--------------------------------------------------------------------------------------------------|
| Preview - auto | Preview - yes | Preview - no |

### DIFF
--- a/modernz.conf
+++ b/modernz.conf
@@ -74,7 +74,7 @@ ytdlp_format=
 
 ## Scaling
 # whether to scale the controller with the video
-vidscale=yes
+vidscale=auto
 # scaling of the controller when windowed
 scalewindowed=1.0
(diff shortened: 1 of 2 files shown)


Samillion:

PR #128, opened 2024-10-30
refactor: change title_mbtn commands to match osc.lua defaults

Keith94:


Samillion:
Awesome. Thank you very much.

PR #135, opened 2024-10-31
refactor: rename osc styles

FreeTheTech101:
Currently, the OSC style variables use a mix of different variable naming conventions, which is most likely due to how many people have worked on the project over the years. This commit standardizes them to use snake care, as well as removes some redundant string concatenation.


Samillion:
You are absolutely right. I only changed a few while I was working on issues. This was a much needed update for consistency. Thank you very much.

Issue #30, opened 2024-10-16
"media-title/no-ext" shows "(error)" if the file has a title tag


Xurdejl:

Samillion:
Damn. I completely forgot about that. Will fix default value, thank you for reporting.

Samillion:
I've set `title` in my [modernz.conf](https://github.com/Samillion/mpv-conf/blob/master/script-opts/modernz.conf) to:
```editorconfig
title=${?demuxer-via-network==no:${filename/no-ext}}${?demuxer-via-network==yes:${media-title}}
```

@Keith94 tagging you as well, since you said before you run with `/no-ext`.

Just leaving a note here, in case it's useful to someone in the future.

Samillion:
The error is at mpv side for sure. Using `media-title` is the preferred method, but I genuinely hate it because there are so many video files with mislabeled titles out there, so I use `filename/no-ext` by default.

The problem is, with `/no-ext` on `media-title`, if the video actually has a title tag, `/no-ext` returns an error because it can't find the extension suffix. This should probably be addressed, I'll mention it to the guys at irc later.

Anywho, this mimics my `Stream` auto-profile, but uses the built in demuxer instead of path check.

```editorconfig
# mpv.conf
# default window title
title=${filename/no-ext}

# for streams, grab media-title
# otherwise it would show url params, ie: ?watch=[ID]
[Stream]
	profile-desc=Apply options when opening a URL or a magnet
	profile-cond=get('path', ''):match('^%a+://') or get('path', ''):match('^magnet:')
	profile-restore=copy
	force-seekable=yes
	title=${media-title}
	save-position-on-quit=no
	user-agent="Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:130.0) Gecko/20100101 Firefox/130.0"
```

Keith94:
>Using media-title is the preferred method, but I genuinely hate it because there are so many video files with mislabeled titles out there

I found this [guess-media-title](https://github.com/zenwarr/mpv-config/blob/master/scripts/guess-media-title.lua) script, which tries to display a nice, clean title output. For example:

- **[Judas] Bleach - Thousand Year Blood War - S17E40.mkv** gets its title displayed as:

<img width="319" height="89" alt="Image" src="https://github.com/user-attachments/assets/55fd95c5-c0d6-4b1c-a535-daa9190e3058" />

[...]

PR #137, opened 2024-10-31
refactor: update title formatting

Keith94:
Makes it match osc.lua defaults.

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -14,6 +14,7 @@ Create `modernz.conf` in your mpv script-opts directory:
 
 | Option         | Value | Description                                              |
 | -------------- | ----- | -------------------------------------------------------- |
+| language       | en    | See [TRANSLATIONS.md](https://github.com/Samillion/ModernZ/blob/main/docs/TRANSLATIONS.md) for other languages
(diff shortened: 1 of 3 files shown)


FreeTheTech101:
Any opinions on defaulting to `${media-title/no-ext}`? Assuming the file in question has a title set, there shouldn't be a difference, but in the event there isn't one and it falls back to the file name, omitting `.mkv` or `.flac` could keep things less cluttered. More so asking to see what everyones opinion is on it, not for any immediate changes.

Keith94:
We used it before but it was removed due to a bug: https://github.com/Samillion/ModernZ/issues/30

Sam provided a workaround for that though.

Samillion:
Awesome! Thank  you very much, I didn't even realize those notes weren't updated.

**Changes to your PR:**
- fix: value for `screenshot_flag`. should be `subtitles`, not `subs`
- fix: correctly close table row with `|` for the `language` option row
- refactor: default title to `${media-title}`. mpv relies on playlist positions because the stock OSC doesn't include native playlist button or tooltips. In ModernZ title would look cluttered with `[2/5] title`. Nevertheless, the user has the option to modify it to their liking.

Keith94:
Thanks for the fix 😃

PR #138, opened 2024-10-31
feat: option to apply hovereffects to slide handles

Commander07:
## Changes
- Added `hovereffect_for_sliders` option which when enabled will apply hovereffects to slide handles (volume & seek bar). This does **not** apply the glow effect as it would not be visible.
## Preview

https://github.com/user-attachments/assets/5f5bc75a-a878-45e3-98e7-656c746fbd0a


Samillion:
Beautiful, thank you very much.

PR #141, opened 2024-11-01
refactor: apply all hover effects by default, except for size, limit it

Samillion:
Pinging: @Commander07 @Keith94 

**Note:** This isn't mandotary, because I realize personal preference is a factor. That is why I decided not to merge but to first consult with you guys. (original PR maker and requester)

The hover effect size looks great on all elements, except for the following: (looks out of place)
- Title (above seekbar)
- Timestamps (left and right)
- Chapter title (below seekbar)

The size hover effect will be stripped from them.

**Other changes:**
- Tie `elem_ass:append()` to a nested condition, to ensure single action. (ie: in title, it was appending twice, hover and text)
- Restore default value for `hover_button_size` to `115` instead of `110`
- Apply all effects in `hovereffect`
- Change `hovereffect_color` to `#CB7050`, a slightly brighter match with the default seekbar color.
- Remove redundant  `or contains(user_opts.hovereffect, "glow")` condition, as it is applied/checked below it already.

**Preview:**

https://github.com/user-attachments/assets/f90439d5-fae3-48b1-b881-4f31a558b721

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -39,6 +39,7 @@ Create `modernz.conf` in your mpv script-opts directory:
 | playpause_color       | #FFFFFF | color of play/pause button                                               |
 | held_element_color    | #999999 | color of an element while held down                                      |
 | thumbnailborder_color | #111111 | color of border for thumbnail (with thumbfast)                           |
+| hovereffect_color     | #CB7050 | color of a hovered button when hovereffect includes: `color`             |
 
 ### Buttons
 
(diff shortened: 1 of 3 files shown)


Keith94:
Yup, the ones you excluded were bothering me the more I used the UI. It's a good change.

The volume handle size change is barely noticeable unless you really squint at it or set a high value like 135

Samillion:
> The volume handle size change is barely noticeable unless you really squint at it or set a high value like 135

😆 true

tenor

Commander07:
Looks much better with these changes.

PR #146, opened 2024-11-02
feat: integrate language locales with a JSON file

Samillion:
Change locales integration to `JSON` instead of `LUA`. This will keep the standard of locale integration while avoiding loading the locales as a script in mpv's backend.

**Changes:**
- Removed `modernz-locale.lua` file
- Added `modernz-locale.json` file
- Added JSON file handler in `modernz.lua` `get_locale_from_json(path)`
- Changed location of `locales` from `scripts/` to `script-opts/`
- Auto prepend ASS tag formatting to `welcome` string (so it doesn't need to be included in JSON)
- Updated main `README.md` to match JSON formatting
- Updated `TRANSLATIONS.md` to match JSON formatting
- Removed unused language labels and phrases
- Changed some language phrases (including translations)

Pinging @Xurdejl to consult


Xurdejl:
Sorry, I've been out all day. The only thing I wanted to ask is if we prefer the locales to be separate files for each language or all in a single file. I suppose, since there aren't many strings, a single file will be good.

Samillion:
Fully agreed. That would definitely be the next stage.

Then maybe match it with a naming scheme like `modernz-LANGCODE.json`, examples:
- `modernz-en.json`
- `modernz-fr.json`
- `modernz-zh.json`

Xurdejl:
That naming scheme sounds perfect, keeps it simple and clear.

PR #150, opened 2024-11-02
fix: consistent punctuation

ctlaltdefeat:


ctlaltdefeat:
Actually already addressed in https://github.com/Samillion/ModernZ/pull/146

PR #151, opened 2024-11-02
feat: add `seekbar_cache_color` option

### LINKED ISSUE #148: Seekbar cache visual indication
By default, `seekbarfg_color` determines both progress (coloring everything before the handle) and cache, meaning that one cannot visually distinguish between cached and non-cached segments before the handle.

This is fine for local content and more closely aligns with the UX of many video players. but it's inconvenient for non-local content.
For myself, I've commented out https://github.com/Samillion/ModernZ/blob/main/modernz.lua#L918, which leads to this:
image

but I think it would be cool to have that be configurable, perhaps as a different color setting?

Commander07:
## Changes
- Added `seekbar_cache_color` which sets the color of the cache ranges. Should close #148.
## Preview
image


Samillion:
As I said in the issue, I never thought about this because I rarely play non-local on mpv. Thank you very much for your effort.

ctlaltdefeat:
Thanks!

PR #152, opened 2024-11-02
feat: add `seek_resets_hidetimeout` option

### LINKED ISSUE #149: Seeking resets hide timer
With `osc_on_seek` set to `no`, seeking does not open the OSC, but it does seem to reset the timeout for hiding it if it was already open.
Given that the description for `hidetimeout` says "duration in ms until OSC hides if no mouse movement", can we make it so that it is indeed determined by mouse movement (or some other configurable way so that seeking does not reset the timer)?

Commander07:
## Changes
- Added `seek_resets_hidetimeout` which controls if seeking should reset the OSC hide timeout. Should close #149.


Samillion:
Thank you very much.

PR #158, opened 2024-11-02
feat: disable irrelevant elements for images

Commander07:
## Changes
- Disables irrelevant elements (play/pause, jump, loop, screenshot, seekbar and time display) for images.
- Removes hover effects from disabled sliders to match button behavior.
- Adds `hide_elements_for_image` which when enabled will hide irrelevant elements for images

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -100,6 +100,7 @@ Create `modernz.conf` in your mpv script-opts directory:
 | playpause_size             | 30            | icon size for the play/pause button                                                                                                                                                                    |
 | midbuttons_size            | 24            | icon size for the middle buttons                                                                                                                                                                       |
 | sidebuttons_size           | 24            | icon size for the side buttons                                                                                                                                                                         |
+| hide_elements_for_image    | yes           | hides irrelevant elements when viewing images                                                                                                                                                          |
 
 ### Colors and style
 
--- a/modernz.conf
+++ b/modernz.conf
@@ -169,6 +169,8 @@ persistentprogress=no
 persistentprogressheight=17
 # on web videos, show the buffer on the persistent progress line
 persistentbuffer=no
+# hides irrelevant elements when viewing images
+hide_elements_for_image=yes
 
 ## UI [behavior]
 # whether to show osc when paused
--- a/modernz.lua
+++ b/modernz.lua
@@ -983,7 +983,7 @@ local function render_elements(master_ass)
                 elem_ass:draw_stop()
                 
                 -- add tooltip
-                if element.slider.tooltipF ~= nil then
+                if element.slider.tooltipF ~= nil and element.enabled then
                     if mouse_hit(element) then
                         local sliderpos = get_slider_value(element)
                         local tooltiplabel = element.slider.tooltipF(sliderpos)
@@ -1863,6 +1863,7 @@ local function osc_init()
     local pl_pos = mp.get_property_number("playlist-pos", 0) + 1
     local have_ch = mp.get_property_number("chapters", 0) > 0
     local loop = mp.get_property("loop-playlist", "no")
+    local is_image = mp.get_property_number("estimated-frame-count", 0) < 2 and audio_track_count == 0
 
     local nojumpoffset = user_opts.showjump and 0 or 100
     local noskipoffset = user_opts.showskip and 0 or 100
@@ -1903,6 +1904,7 @@ local function osc_init()
     --play control buttons
     --play_pause
     ne = new_element("play_pause", "button")
+    ne.enabled = not is_image
     ne.content = function ()
         if mp.get_property("eof-reached") == "yes" then
             return icons.replay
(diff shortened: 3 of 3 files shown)


Samillion:
There is no freakin way. I was JUST talking to mpv maintainers about implementing this very same thing. 😆 

image

Samillion:
I like your implementation of checking frame count. Mind if I hold on merging for a bit? In theory, you can check for image in a more accurate way, to ensure consistency.

I'm in a very lazy mood so sticking in bed for a while lol.

An example of why checking frame count only might be a problem: audio files with coverart. Having additional checks to `is_image` would prevent it, ie: `current-tracks/video/image`

Samillion:
Leaving this as a note, alongside frame count.

```lua
local current_track = mp.get_property_native("current-tracks/video")
local is_image = current_track and current_track.image and not current_track.albumart and mp.get_property_native("estimated-frame-count", 0) < 2
```

Commander07:
> I like your implementation of checking frame count. Mind if I hold on merging for a bit? In theory, you can check for image in a more accurate way, to ensure consistency.
> 
> I'm in a very lazy mood so sticking in bed for a while lol.
> 
> An example of why checking frame count only might be a problem: audio files with coverart. Having additional checks to `is_image` would prevent it, ie: `current-tracks/video/image`

There is a check for audio tracks in the `is_image` check which makes sure audio files don't trigger it. However more checks won't hurt. I don't have time to do that today, so I'll look into it tomorrow.

Side note: Should there be an option to hide the seekbar and the other non regular media controls (play/pause, next/prev, jump and fullscreen) instead of disabling/graying out them?

Samillion:
> There is a check for audio tracks in the is_image check which makes sure audio files don't trigger it. However more checks won't hurt. I don't have time to do that today, so I'll look into it tomorrow.

Yeah, I noticed and edited my reply right before you replied lol.

> Side note: Should there be an option to hide the seekbar and the other non regular media controls (play/pause, next/prev, jump and fullscreen) instead of disabling/graying out them?

Definitely. For my use case, I'd only want the relevant buttons showing.

Samillion:
I realize the additional checks might be an overkill, but it's worth testing to be honest. To bulletproof `is_image` state. So it's not a replacement for your check, it's addition to it.

I'll test more once I get the energy to out of bed.

Samillion:
> Side note: Should there be an option to hide the seekbar and the other non regular media controls (play/pause, next/prev, jump and fullscreen) instead of disabling/graying out them?

Your side note made me realize a scenario. playlist track next/previous buttons should be forced to show, regardless of `user_opts.shownextprev`, and if `have_pl`, right?

I think out of all the media control buttons, those two are the only ones that would make sense to show them (if it's a playlist)

What do you think?

Commander07:
> Your side note made me realize a scenario. playlist track next/previous buttons should be forced to show, regardless of `user_opts.shownextprev`, and if `have_pl`, right?
> 
> I think out of all the media control buttons, those two are the only ones that would make sense to show them (if it's a playlist)
> 
> What do you think?

Because the playlist button will show up regardless if next/prev is enabled, so i think it's safe to allow next/prev to be hidden along with the rest of the buttons.

Hiding enabled, next/prev enabled:
image

Hiding enabled, next/prev disabled:
image


I have also encountered a weird bug where the download button is not being shown when `showjump` and `showskip` are both true and I'm not sure whats causing it. Could you have a look and see if you can spot the issue?

Samillion:
Absolutely beautiful work. Love the placement of title. Do long titles overlap the buttons?

> I have also encountered a weird bug where the download button is not being shown when showjump and showskip are both true and I'm not sure whats causing it. Could you have a look and see if you can spot the issue?

Download button only shows if it's a web video (ie: path is url)
https://github.com/Samillion/ModernZ/blob/c053c5540bfc3c60301fce4e1bd2ad5de0856ed8/modernz.lua#L1351-L1354

https://github.com/Samillion/ModernZ/blob/c053c5540bfc3c60301fce4e1bd2ad5de0856ed8/modernz.lua#L2175-L2178

Samillion:
By the way, if the title does overlap, you can invoke `maxchars`, like so:
https://github.com/Samillion/ModernZ/blob/c053c5540bfc3c60301fce4e1bd2ad5de0856ed8/modernz.lua#L1528

Commander07:
> > I have also encountered a weird bug where the download button is not being shown when showjump and showskip are both true and I'm not sure whats causing it. Could you have a look and see if you can spot the issue?
> 
> Download button only shows if it's a web video (ie: path is url)

Just tried it now and it works i must have accidentally tested on a local file.

Some clarifications on some perhaps unclear communication from me:
- When next/prev is enabled and there is no playlist they will be grayed out.
image
- Playlist button and next/prev can both be disabled
image

So unless there are any changes or additions i feel that this PR is complete and ready for testing so it can be merged.

Samillion:
Great effort, thank you so much. Merging.

PR #159, opened 2024-11-03
docs: reorder user options and improve comments for clarity

Xurdejl:
Initially I was going to reword some comments for clarity but ended up reordering the user options as well. The reordering reflects my personal preference on what I think makes sense, so I'm open to feedback on whether this approach is effective or if there are better ways to structure it.

### Changes
- Updated the header comment: Added licensing information and provided some info, aligning it with standard practices seen in other scripts.
- Made some comments more clear for better understanding, while trying to be consistent.
- Reordered and regrouped the user options.
- Changed the "log" value to "logarithmic" for consistency.

I considered splitting the buttons group into two distinct groups: one for showing and hiding buttons and another for the button functionality settings. It would look something like this:

``` lua
    -- Button visibility
    showjump = true,                       -- show "jump forward/backward 10 seconds" buttons
    showskip = false,                      -- show the skip back and forward (chapter) buttons
    shownextprev = true,                   -- show next/previous playlist track buttons
    volumecontrol = true,                  -- show mute button and volume slider
    showplaylist = false,                  -- show playlist button
    downloadbutton = true,                 -- show download button on web videos (requires yt-dlp and ffmpeg)
    showscreenshot = false,                -- show screenshot button
    showontop = true,                      -- show window on top button
    showloop = true,                       -- show loop button
    showinfo = false,                      -- show info button
    showfullscreen_button = true,          -- show fullscreen toggle button

    -- Button functionality
    jumpamount = 10,                       -- jump amount in seconds
    jumpiconnumber = true,                 -- show different icon for 5, 10, or 30 second jumps
    jumpmode = "relative",                 -- seek mode for jump buttons
    jump_softrepeat = true,                -- enable continuous jumping when holding down seek buttons
    chapter_softrepeat = true,             -- enable continuous skipping when holding down chapter skip buttons
    volumecontroltype = "linear",          -- volume scale type: "linear" or "logarithmic"
    hide_empty_playlist_button = true,     -- hide playlist button when no playlist exists
    gray_empty_playlist_button = true,     -- gray out the playlist button when no playlist exists
	download_path = "~~desktop/mpv",       -- default download directory for videos (https://mpv.io/manual/master/#paths)
    screenshot_flag = "subtitles",         -- flag for screenshot button
    loopinpause = true,                    -- enable looping by right-clicking pause
````

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,11 +35,12 @@ Don't like the default white buttons and text? ModernZ is fully customizable! Ch
 
 ![modernz-colors-bottom](https://github.com/user-attachments/assets/6d686c53-843b-46f9-b630-a396e88b95be)
 
-See the [Color Customization](docs/USER_OPTS.md#colors) section in the configuration guide for details on how to customize colors and buttons.
+See the [Color Customization](docs/USER_OPTS.md#colors-and-style) section in the configuration guide for details on how to customize colors and buttons.
 
 ## Installation
 
 1. **Disable Stock OSC**
+
    - Add `osc=no` in your `mpv.conf`
 
 2. **Copy Files**
@@ -72,7 +73,7 @@ See the [Color Customization](docs/USER_OPTS.md#colors) section in the configura
 
 1. Create `modernz.conf` in the `/script-opts` folder to customize settings
 
-   - [Download default modernz.conf](https://github.com/Samillion/ModernZ/blob/main/modernz.conf)
+   - [Download default modernz.conf](./modernz.conf)
 
 2. Example short configuration:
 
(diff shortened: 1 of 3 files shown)


Samillion:
Brilliant. Especially because there have been many copyright and credit considerations happening not just mpv side, but libraries used as well. It was all started by ImageMagick.

**References:**
https://code.videolan.org/videolan/libplacebo/-/merge_requests/679
https://code.videolan.org/videolan/libplacebo/-/issues/329
https://github.com/mpv-player/mpv/pull/15083
https://github.com/mpv-player/mpv/pull/15131

Your PR covers all the bases that we would ever need.

> I considered splitting the buttons group into two distinct groups: one for showing and hiding buttons and another for the button functionality settings.

That looks perfect. We should do these changes to the [user options page](https://github.com/Samillion/ModernZ/blob/main/docs/USER_OPTS.md) in the future as well

Samillion:
Oh, holy crap. You did the user options page as well. Thank you so much, they look perfect.

Xurdejl:
> Oh, holy crap. You did the user options page as well. Thank you so much, they look perfect.

I'm sure there's a way to automate it with gitHub actions, but I'm not sure if it's worth the extra setup here.

Samillion:
Definitely doable with a regex match, but I think at one point the user options will stop being edited or rarely so.

The next big change, which I'm scared of, is renaming the whole `user_opts` so that everything matches or makes sense.

After these stages, we'll rarely see additions. It is already such a long list. 😆

Xurdejl:
> The next big change, which I'm scared of, is renaming the whole `user_opts` so that everything matches or makes sense.

I can help with that. What naming conventions do we want to follow for the keys?
I suppose we should also follow some convention for comments like start with "set", "show", etc...

Samillion:
> I can help with that. What naming conventions do we want to follow for the keys? I suppose we should also follow some convention for comments like start with "set", "show", etc...

Done in: https://github.com/Samillion/ModernZ/pull/170

PR #161, opened 2024-11-03
fix: correctly fill cache color in slider

Samillion:
Pinging @Commander07 and @ctlaltdefeat

There is a gap at the end of the seekbar, because the cache color doesn't address the handle size

| Before | After |
|--------|-------|
| Before | After |

**Test with:** 
- [dev_slider_gap_fill/modernz.lua](https://github.com/Samillion/ModernZ/blob/dev_slider_gap_fill/modernz.lua)

**Changes:**
```diff
- elem_ass:rect_cw(pstart, slider_lo.gap, pend, elem_geo.h - slider_lo.gap)
+ elem_ass:rect_cw(pstart - rh, slider_lo.gap, pend + rh, elem_geo.h - slider_lo.gap)
```

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -969,8 +969,8 @@ local function render_elements(master_ass)
                         local pend = get_slider_ele_pos_for(element, range["end"])
 
                         local cache_starts_in_handle = pstart >= xp-rh and pstart <= xp + rh
-                        local cache_ends_in_handle = pend >= xp-rh and pend <= xp + rh
-                        local cache_passes_handle = pstart < xp-rh and pend > xp + rh
+                        local cache_ends_in_handle = pend >= xp-rh and pend <= xp
+                        local cache_passes_handle = pstart < xp-rh and pend > xp
                         if cache_starts_in_handle or cache_ends_in_handle then
                             if cache_starts_in_handle and cache_ends_in_handle then
                                 pstart = 0
@@ -988,7 +988,7 @@ local function render_elements(master_ass)
                             pstart = pstart - rh
                         end
 
-                        elem_ass:rect_cw(pstart, slider_lo.gap, pend + rh, elem_geo.h - slider_lo.gap)
+                        elem_ass:rect_cw(pstart, slider_lo.gap, pend + (cache_ends_in_handle and 0 or rh), elem_geo.h - slider_lo.gap)
                     end
                 end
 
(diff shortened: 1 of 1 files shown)


Commander07:
`pstart - rh` will cause part of the cache to render above the handle.
image

Commander07:
image
One more issue: A cache range at the start of the video which does not pass or start/end in the handle will have a small gap.
Should be solvable by adding `pstart = pstart - rh` in an else branch [here](https://github.com/Samillion/ModernZ/pull/161/commits/c36726575e77dcee845f42bf8e1ccf38eb7c2468#diff-d5f739bcfd17a07337246b9195b4fd31e4624b22622aaf2e8e8265f4820789b1R985).

Samillion:
Thank you very much for the notes, I've applied them. Should be good now. If I missed anything, please let me know.

Commander07:
> Thank you very much for the notes, I've applied them. Should be good now. If I missed anything, please let me know.

There seems to be another edge case. When the handle is nearing the end of the seekbar the cache slightly renders above it and cuts out the last bit of the cache.
image

<details><summary>Notes</summary>
The first issue seems to be fixed by changing line 991:

```diff
- elem_ass:rect_cw(pstart, slider_lo.gap, pend + rh, elem_geo.h - slider_lo.gap)
+ elem_ass:rect_cw(pstart, slider_lo.gap, pend + (cache_ends_in_handle and 0 or rh), elem_geo.h - slider_lo.gap)
```
Resulting in:
image

The second issue seems to be caused by `cache_ends_in_handle` incorrectly being set to true. Which seems to be solved changing lines 972-973:
```diff
- local cache_ends_in_handle = pend >= xp-rh and pend <= xp + rh
- local cache_passes_handle = pstart < xp-rh and pend > xp + rh
+ local cache_ends_in_handle = pend >= xp-rh and pend <= xp
+ local cache_passes_handle = pstart < xp-rh and pend > xp
```
This does however result in `cache_passes_handle` being set to true when the cache ends within the right half of the seekbar handle. But the second draw done by `cache_passes_handle` is called with `pstart = xp+rh` and `pend = xp+rh` so nothing is drawn.
However, the `pstart` value of the first draw in this scenario should result in a draw over of `xp-pend` but works fine without correction but does draw over with correction. So i suspect i missed something either in implementation or testing.
image
</details>

Samillion:
I've applied your notes, it seems to be working for me, but we can fix any issues later.

I'm opening another PR that is kind of a major change, an addition to your image mode, well, an adjustment.

I'll merge this one, make the other PR, then check everything there.

PR #166, opened 2024-11-03
fix: `osc_on_seek` not triggering for backwards seeking

### LINKED ISSUE #164: osc_on_seek doesn't show OSC when seeking backwards
Seeking forward is OK but backwards does not show_osc()

Maybe caused by https://github.com/Samillion/ModernZ/commit/3cb9293f0bd1361a061122ae8de92f5182fc05ec

Commander07:
Changes the way `new_file_flag` is handled to avoid it being triggered on backwards seeking. Should close #164.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -2985,8 +2985,9 @@ local function set_tick_delay(_, display_fps)
     end
     tick_delay = 1 / display_fps
 end
-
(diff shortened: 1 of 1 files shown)


Samillion:
Much simpler logic, thank you.

PR #167, opened 2024-11-03
fix: `loopinpause` not being checked

### LINKED ISSUE #165: loopinpause boolean does nothing
Right click on pause will loop no matter what. Seems like there's no checks for user opt `loopinpause`

https://github.com/Samillion/ModernZ/blob/9b0e667d864fb41dbfcc39c53e324d5ed03ae637/modernz.lua#L1959-L1967

Commander07:
Checks if `loopinpause` when right clicking play/pause button. Should close #165.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -1957,13 +1957,15 @@ local function osc_init()
         end
     end
     ne.eventresponder["mbtn_right_down"] = function ()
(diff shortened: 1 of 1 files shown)


Samillion:
Thank you very much.

PR #174, opened 2024-11-04
refactor: move slider seekranges & handle drawing into own functions

### LINKED ISSUE #173: seekbar cache color not applied to persistent buffer
It seems that these features both show the buffer status of web videos, but `persistentbuffer` line doesn't follow the same color.

Commander07:
This PR aims to reduce code duplication and complexity by separating slider seekrange & handle drawing from the rest of the slider. As a result seekrange drawing is now the same on both the normal and persistent seekbar which closes #173.
## Changes
- Created `draw_slider_handle` function which takes an element its ass and draws a handle using settings provided by `user_opts`.
- Created `draw_slider_seekranges` function which takes an element its ass and draws seekranges using settings provided by `user_opts`.
- Changed both sliders and the persistent seekbar to use the new functions,
## Preview
Persistent seekbar now respects `user_opts.seekbar_cache_color`
image

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -870,6 +870,85 @@ local function get_chapter(possec)
     end
 end
 
+-- Draws a handle on the slider according to user_opts
+-- Returns handle posistion and radius
+local function draw_slider_handle(element, elem_ass, override_alpha)
+    local pos = element.slider.posF()
+    if not pos then
+        return 0, 0
+    end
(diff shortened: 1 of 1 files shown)


Samillion:
Absolutely beautiful. I think at this point you've improved most of the seekrange behavior on ModernZ. Much appreciated.

PR #177, opened 2024-11-05
feat: made the secondary seek amounts configurable

Keith94:
Made the seek commands configurable, so they're no longer forced to 60 seconds.


Samillion:

PR #233, opened 2024-11-17
Feat: make slider handle hover-only with user option

Xurdejl:
This PR introduces a new user option, `handle_always_visible`, to control the visibility of the slider handles (progress bar, volume, etc.).

- When `handle_always_visible` is set to `"true"`, the slider handle remains visible at all times, following the previous behaviour.
- When `handle_always_visible` is set to `"false"`, the handle will only be visible when the user hovers over the slider, just like YouTube, Spotify, etc.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -931,11 +931,10 @@ local function draw_slider_handle(element, elem_ass, override_alpha)
 end
 
 -- Draws slider seekranges according to user_opts 
-local function draw_slider_seekranges(element, elem_ass, xp, rh, override_alpha)
+local function draw_slider_seek_ranges(element, elem_ass, xp, rh, override_alpha)
     rh = rh or 0
-    local handle = true
+    local handle = xp and user_opts.handle_always_visible
     if not xp then
-        handle = false
         xp = 0
     end
     local slider_lo = element.layout.slider
@@ -1042,7 +1041,7 @@ local function render_elements(master_ass)
 
                 local xp, rh = draw_slider_handle(element, elem_ass) -- handle posistion, handle radius
                 elem_ass:rect_cw(0, slider_lo.gap, xp-rh, elem_geo.h - slider_lo.gap)
-                draw_slider_seekranges(element, elem_ass, xp, rh)
+                draw_slider_seek_ranges(element, elem_ass, xp, rh)
 
                 elem_ass:draw_stop()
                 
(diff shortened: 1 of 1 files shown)


Samillion:
Pinging @Commander07

Keith94:
Great addition for sure.

When you hover any part of the seekbar, would be nice  to have an animation where the handle enlarges from 0 and shrinks back to 0 when not hovered (see YouTube)

Commander07:
> One issue, this would create a gap where the slider handle would be (outside hover)

Should be easily fixable by setting `handle` inside of `draw_slider_seekranges` according to the same logic as `should_display_handle` here:
https://github.com/Samillion/ModernZ/blob/b0c7bb0d825f7f614bd1fa222f87b6722baf57a6/modernz.lua#L918-L924

Side note: `draw_slider_seekranges` should probably be renamed to `draw_slider_seek_ranges`

Samillion:
I apologize again for the late response and lack of testing. I could be completely wrong, because I'm making this comment based on looking at the code only without testing. Please let me know.

You can test this easily if you open a video, seek to near end. You'll see that there is a seek gap (not cache). That happens because it's where the seek handle is supposed to be, basically.

The way it works is `===seekbar==[handle]` where `[handle]` is a completion of the seek bar.

Since now handle is optional (on hover), the seek bar is: `===seekbar==[gap]---[cache]` (no hover), the gap is filled with the seek handle on hover.

If I may suggest a solution: making seek bar not account seek handle size as part of it. In essence, turning seek handle into a hover effect that replaces the last `=`, if that makes any sense.

- `===[seekbar]===---[cache]` => no hover
- `===[seekbar]==[handle]---[cache]` => no hover

Commander07:
I got carried away with testing and came up with a solution to deal with the position and size inaccuracies of the seekbar and ranges. It's still a bit messy but seems to work mostly fine. The solution Sam suggested is probably better but I'm putting this here as an alternative.

https://github.com/Xurdejl/ModernZ/pull/1

Samillion:
> The solution Sam suggested is probably better but I'm putting this here as an alternative.

Thank you very much for your effort in this, honestly I think your implementation is better because after thinking about it, this feature update allow users to decide whether they want the handle like the usual modern osc behavior or this new efficient one.

Once @Xurdejl merges and pushes your PR and you guys are satisfied, I'll merge.

I know I sound like a broken record, but again, terribly sorry for my absence in testing. I drank a liter of tea with lemon and honey to force my body to feel better lol. I'll come back in full force very soon. I promise.

Samillion:
Code looks perfect. @Xurdejl @Commander07 Should I merge?

Can't wait to use this, genuinely a great idea.

Commander07:
> Code looks perfect. @Xurdejl @Commander07 Should I merge?

I have no issues with this being merged in its current state.

Xurdejl:
yes, ready to merge

PR #242, opened 2024-11-21
fix: improve consistency for width and coordinate values

Keith94:
Improves the hit boxes for time stamps to be more accurate, and adjusts the chapter title separator to be more consistent when time stamp width changed and for the tooltip.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -1149,7 +1149,7 @@ local function render_elements(master_ass)
                                 if dur > 0 then
                                     local ch = get_chapter(state.sliderpos * dur / 100)
                                     if ch and ch.title and ch.title ~= "" then
-                                        tooltip_content = tooltip_content .. " • " .. string.format(user_opts.chapter_fmt, ch.title)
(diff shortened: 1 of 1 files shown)


Samillion:
Thank you very much for your efforts.

By any chance you're using a custom font for mpv? I think that's why we're seeing different results.

image

Keith94:
Actually, yeah I am. Oops. I will test again.

Keith94:
Crap. The font was messing things up. I was using Open Sans Semi Bold.

Everything looks ok with default font. Nevermind lol.

Samillion:
It honestly brings up a good point. One of the main reasons why I hate OSC being libass based. It is essentially tricking the system into displaying text as elements.

Which makes something as simple as changing font type a breaking change.

I agree with you that the default font isn't good, but I can't think of a good way to use a better replacement that wouldn't apply workload on the users.

Maybe one day we should discuss this in detail.

Keith94:
Yeah, I had no idea the geometry was that delicate until you mentioned custom font.

On one hand, it's nice that people can choose their own font, but it's very much "at your own risk" because of the breakage.

We would have to either support a new, specific font type, which would break a lot or configurations, or simply leave it as is.

PR #247, opened 2024-11-23
fix: respect OSC visibility over paused visibility

### LINKED ISSUE #246: OSC shown on pause when visibility=never
### mpv Information

_No response_

### Reproduction Steps

1. Set `visibility = "never",` and `showonpause = true,`
2. Open a file and pause playback.

### Expected Behavior

OSC not shown

### Actual Behavior

It shows

### Log File

_No response_

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information.

FreeTheTech101:
Adds a conditional check to `showonpaused` so it won't trigger when the user has the visibility set to `never`. Closes #246.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -3378,7 +3378,7 @@ end
 
 mp.observe_property("pause", "bool", function(name, enabled)
     pause_state(name, enabled)
-    if user_opts.showonpause then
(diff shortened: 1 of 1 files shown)


Samillion:
Perfect. Thank you very much.

PR #269, opened 2024-12-24
feat: add additional mouse functions for buttons

Keith94:
Based on https://github.com/mpv-player/mpv/blob/master/DOCS/man/osc.rst


Samillion:
Awesome. Thank you very much.

Issue #239, opened 2024-11-21
If handle_always_visible is disabled, the start and end of progress are not displayed correctly


Keith94:
### Reproduction Steps

See https://github.com/Samillion/ModernZ/issues/239#issuecomment-2565767127

Samillion:
CC @Commander07

Samillion:
@Xurdejl terribly sorry, forgot to CC you as well.

Keith94:
I came across an interesting line dealing with `progress_offset`:

https://github.com/Samillion/ModernZ/blob/06cc6515f0ff000a988cb5a735d914b0aec91991/modernz.lua#L988

So I deleted `slider_lo.gap`, which fixed that weird bug mentioned above.

```lua
local progress_offset = (pos / 100) * math.abs(pos / 100 - 1)
```

I also deleted this code because it seems not needed anymore since values of 0 are lined up now.

https://github.com/Samillion/ModernZ/blob/06cc6515f0ff000a988cb5a735d914b0aec91991/modernz.lua#L3437-L3440

However, I found another issue, where the handle size was affecting the position of the chapter nibbles. Bigger values like 1 caused the min and max points to be pushed in closer.

![image](https://github.com/user-attachments/assets/bcd44c62-2936-4eb1-bf83-fef3ac906fe3)

Samillion:
> So I deleted `slider_lo.gap`, which fixed that weird bug mentioned above.

😱 y u no make pr?

Keith94:
Welp. I didnt author that code so maybe there was a good reason for it being there. As such there could be a better way to fix? Wasn't sure.

I could make a PR later.

Samillion:
It does indeed improve, but absolute start and end are incorrect with fill.

Keith94:
I guess [7486cbc](https://github.com/Samillion/ModernZ/commit/7486cbcb88bef31f27150091344578e747e3606c) should be reverted until a better workaround is found imo..

Samillion:
> I guess [7486cbc](https://github.com/Samillion/ModernZ/commit/7486cbcb88bef31f27150091344578e747e3606c) should be reverted until a better workaround is found imo..

It's only broken if `handle_always_visible` is set to `no/false`, right?

Should be fine as the default behavior is `yes/true`. Most likely it just needs adjusting in the ranges function for `xp` and `rh` values. I hate math, it makes my tiny brain hurt, that is why I'm delaying looking into it. 😝

[...]

PR #277, opened 2024-12-29
feat: add more chapter marker styles with nibbles_style

xiaoleichen:
Two styles:

* existing triangle style

Screenshot 2024-12-28 at 3 19 09 PM

* More subtle bar style

Screenshot 2024-12-28 at 3 17 45 PM

Also slider gap should be computed from seekbar background height and seekbar height instead of redefining.

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -156,21 +156,22 @@ Create `modernz.conf` in your mpv script-opts directory:
 
 ### Progress bar settings
 
-| Option                   | Value | Description                                                                |
-| ------------------------ | ----- | -------------------------------------------------------------------------- |
-| seekbarhandlesize        | 0.8   | size ratio of the seekbar handle (range: 0 ~ 1)                            |
-| handle_always_visible    | no    | control handle visibility: `yes` always visible, `no` show on slider hover |
-| seekrange                | yes   | show seek range overlay                                                    |
-| seekrangealpha           | 150   | transparency of the seek range                                             |
-| livemarkers              | yes   | update chapter markers on the seekbar when duration changes                |
-| seekbarkeyframes         | no    | use keyframes when dragging the seekbar                                    |
-| nibbles_top              | yes   | top chapter nibbles above seekbar                                          |
-| nibbles_bottom           | yes   | bottom chapter nibbles below seekbar                                       |
-| automatickeyframemode    | yes   | automatically set keyframes for the seekbar based on video length          |
-| automatickeyframelimit   | 600   | videos longer than this (in seconds) will have keyframes on the seekbar    |
-| persistentprogress       | no    | always show a small progress line at the bottom of the screen              |
-| persistentprogressheight | 17    | height of the persistent progress bar                                      |
-| persistentbuffer         | no    | show buffer status on web videos in the persistent progress line           |
+| Option                   | Value    | Description                                                                |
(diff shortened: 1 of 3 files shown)


Keith94:
What about making the bars transparent similar to YouTube's player?

Nibbles follow the seekbar color (mentioned in https://github.com/Samillion/ModernZ/discussions/243) but maybe we can override it.

image

xiaoleichen:
> What about making the bars transparent similar to YouTube's player?
> 
> Nibbles follow the seekbar color (mentioned in #243) but maybe we can override it.

I don't think setting nibbles transparent will be able to break down the seekbar as Youtube does. Maybe setting a different color from the seekbar would make it feel like so. Alternatively we can draw seekbar itself in segments, which doesn't feel very straightforward to implement.

Samillion:
Beautiful work, thank you very much. Though if I may, I have an input I'd like to share. To be exact, I 100% know that mpv users will request this adjustment.

The rectangle marker splits through the seek bar, as it is drawn as one element. Why not follow the triangle nibbles way and make it top/bottom thin squares or rectangles?

This would:
- Ensure consistent user experience and future maintenance
- Be compatible with current user options (`nibbles_top`, `nibbles_bottom`)

xiaoleichen:
> The rectangle marker splits through the seek bar, as it is drawn as one element. Why not follow the triangle nibbles way and make it top/bottom thin squares or rectangles?

That was how I initially did it, but I didn't really like it. I thought since the width of the markers is vertically consistent now, I like it better to cut through the seekbar.

This is the initial implementation as you suggested:
Screenshot 2024-12-29 at 12 01 36 AM

Also current implementation doesn't draw the bar as one element. It still draws top and down nibbles separately, and both bars extend through the seekbar to make it look like one. It still supports top or bottom only, though I'll admit it doesn't look as good when showing top / bottom only:
Screenshot 2024-12-29 at 12 21 41 AM


If you think people would prefer the initial implementation with separate top/bottom rectangles, I can revert to it, or make them separate styles. Either way me personally like both of them better than triangles which can be too distracting, especially when there are many chapters in a video.

Samillion:
> If you think people would prefer the initial implementation with separate top/bottom rectangles, I can revert to it, or make them separate styles. Either way me personally like both of them better than triangles which can be too distracting, especially when there are many chapters in a video.

It is unfortunately one of the downsides of trying to please everyone. I genuinely agree with you, but over the last 6 years I've seen enough reports from users to know how sometimes even the smallest change can be contested.

A good example on mpv repo itself, it took months and over 90+ comments (and one closed PR), to agree on a font size change for console. I am in no way saying it wasn't a valid discussion, I'm merely pointing out that I know this will be requested at one point or the other.

It is why I generally tend to do personal preference as optional as possible.

In this case, the two nibbles behavior. If you want to keep it strike through, then it needs to be added as a third style option, if that makes sense.

Otherwise, you'll be facing many requests for adjustments, and as we can see in #239 a delay or abandoned status can easily happen.

xiaoleichen:
Makes sense to me!

Added bar and single-bar styles to implement the two styles shown in above screenshots.

Samillion:
Absolutely perfect. Thank you very much for your effort and open mind in regards to the adjustments.

Will merge now.

I can already feel the feature request being made "make it so we can control nibbles color!" 😆

PR #294, opened 2025-01-05
fix: don't hide time codes in portrait or narrow window

Samillion:
Instead of hiding the time codes, move them above play controls, centered.

Centered above play controls is the best fit, as these buttons do not have tooltips and places the essential elements centered so even in abnormal narrow videos or window, they're displayed.

https://github.com/user-attachments/assets/07cb8d32-06b8-44e3-93e9-d13ce5888b8d

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -1830,6 +1830,7 @@ layouts["modern"] = function ()
     local show_remhours = (state.tc_right_rem and remsec >= 3600) or (not state.tc_right_rem and dur >= 3600) or user_opts.time_format ~= "dynamic"
     local tc_w_offset = (state.tc_ms and 60 or 0) + (show_hours and 20 or 0) + (show_remhours and 20 or 0)
     local auto_hide_volbar = (audio_track and user_opts.volume_control) and osc_param.playresx < (1150 - outeroffset)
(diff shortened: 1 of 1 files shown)


Samillion:
CC: @Keith94 @BODZy @jiangzhenjerry
Sorry for the random ping, looking for feedback on this PR before I merge, Only if it's not a bother.

**Test script**: [portrait_videos_tc/modernz.lua](https://github.com/Samillion/ModernZ/blob/dev_portrait_timecodes/modernz.lua)

Keith94:
That's an awesome idea. I definitely approve.

You might see a clash/overlap of elements with larger font sizes (such as 18) or when milliseconds are enabled.

image

Samillion:
True! I did not think of that. I'll apply a fix for that one in a different PR, thank you so much for pointing it out. Completely slipped my mind.

jiangzhenjerry:
Thanks for pinging. So far so good on my end. `time_font_size` >=20 may make it look a little crowded for portrait videos but 18 and below works well. Love it.

Samillion:
Awesome, thank you for confirming. I'll apply a PR fix for time state and font size after merging this one.

Samillion:
Merging. Follow up on the other issues here: https://github.com/Samillion/ModernZ/issues/295

Issue #306, opened 2025-01-11
Always show seekbar tooltip info while dragging


Keith94:
### Current behavior

1. Put bottomhover=no
2. Click and hold seekbar handle
3. Move mouse above the seekbar
4. Move mouse left/right

All seekbar information (thumbnail, time stamp, chapter title) disappears unless you're hovering the seek handle.

### Desired behavior

Keep all seekbar information prominent even if mouse cursor drifts off the handle while dragging.

Samillion:
In theory, it is possible, just need to modify the mouse hit to add a condition to it.

https://github.com/Samillion/ModernZ/blob/488854bb2d282c3486fb05169552043ad9b758f6/modernz.lua#L1138-L1143

For example, if we add this before that part of code:
```lua
local force_show_tooltip = element.name == "seekbar" and 
    element.eventresponder["mbtn_left_down"] 
    and state.mouse_down_counter > 0 
```

Then replace `if mouse_hit(element) then` to:
```lua
if mouse_hit(element) or force_show_tooltip then
```

**Issues**:
- Not tested
- Most likely will show thumbnails with volume bar as well
- `force_show_tooltip` most likely needs better identifiers to specifically and only work for seekbar, not all sliders
- There is probably an obvious modifier that my brain either forgot or is not a real enough brain to think of lol

Keith94:
Awesome I can definitely test it out as soon as I can. Let you know if any issues pop up.

Keith94:
It mostly works well (way better than current behavior) but with a couple glitches like the one you mentioned:

>Most likely will show thumbnails with volume bar as well

There's another glitch, if you drag the progress bar to the beginning or end of the file, click and drag the maximize button, the thumbnail will show up.

https://github.com/user-attachments/assets/ac3df4fb-b15a-4a95-9c26-cd7f9dff6188

Samillion:
Try:
```lua
                    local force_show_tooltip = element.name == "seekbar"
                        and element.eventresponder["mbtn_left_down"]
                        and element.state.mbtnleft
                        and state.mouse_down_counter > 0
                        and state.playingWhilstSeeking
```

Keith94:
I couldn't get it working; once I added `and state.playingWhilstSeeking`, the time and thumbnails don't show up unless hovered.

Samillion:
[...]

PR #338, opened 2025-02-01
feat: add force_seek_tooltip option

Samillion:
Fixes: https://github.com/Samillion/ModernZ/issues/306

Force show seekbar tooltip (ie: thumbnail) on mouse drag, even if not hovering seekbar, while holding down mouse button.

### Changes
- Add `force_seek_tooltip` option
- Rename `playingWhilstSeeking` state to `playing_and_seeking`
- Rename `persistentprogresstoggle` state to `persistent_progress_toggle`
- Remove unused state `playingWhilstSeekingWaitingForEnd`
- Make `playing_and_seeking` state independent and not tied to `mouse_seek_pause` option
- Fix behavior of  `mouse_seek_pause` option when set to `no` and mouse drag then release mouse button
  - It used to pause the video on button release, now it doesn't
  - Now it only cycles pause if `mouse_seek_pause` is set to `yes`

### Usage
```EditorConfig
# modernz.conf
# enable the option to force seek tooltip
force_seek_tooltip=yes
```

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -26,19 +26,20 @@ Create `modernz.conf` in your mpv script-opts directory:
 
 ### OSC behaviour and scaling
 
-| Option                  | Value | Description                                                |
-| ----------------------- | ----- | ---------------------------------------------------------- |
-| hidetimeout             | 2000  | time (in ms) before OSC hides if no mouse movement         |
-| seek_resets_hidetimeout | yes   | if seeking should reset the hidetimeout                    |
-| fadeduration            | 250   | fade-out duration (in ms), set to `"0"` for no fade        |
-| minmousemove            | 0     | minimum mouse movement (in pixels) required to show OSC    |
-| bottomhover             | yes   | show OSC only when hovering at the bottom                  |
-| bottomhover_zone        | 130   | height of hover zone for bottomhover (in pixels)           |
-| osc_on_seek             | no    | show OSC when seeking                                      |
-| mouse_seek_pause        | yes   | pause video while seeking with mouse move (on button hold) |
-| vidscale                | auto  | scale osc with the video. (set to `"no"` to disable)       |
-| scalewindowed           | 1.0   | osc scale factor when windowed                             |
-| scalefullscreen         | 1.0   | osc scale factor when fullscreen                           |
+| Option                  | Value | Description                                                             |
+| ----------------------- | ----- | ----------------------------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


Keith94:
There's some minor tweaks you could potentially add to this.

1. Disable tooltips while dragging seekbar: if you're seeking and move cursor down to any button, it makes more sense to hide all button tooltips (since you can't interact while dragging anyway).

2. Disable `hover_effect_for_sliders` while seeking: hovering either handle during seekbar drag triggers the hover size effect, despite the hover color effects being disabled on all other buttons.

Samillion:
> 2\. Disable `hover_effect_for_sliders` while seeking: hovering either handle during seekbar drag triggers the hover size effect, despite the hover color effects being disabled on all other buttons.

Could you elaborate? size effect and color effect are different. You said "despite all hover color effects being disabled". Disabling  color doesn't affect sliders, as sliders only have size effects applied to them, nothing else.
```lua
        -- Apply size hover_effect only if hovering over the handle
        if handle_hovered and user_opts.hover_effect_for_sliders then
            if contains(user_opts.hover_effect, "size") then
                rh = rh * (user_opts.hover_button_size / 100)
            end
        end
```

Samillion:
> My bad, I phrased that poorly.

Not at all, English is not my first language, sometimes my brain scrambles trying to understand simple things xD, my apologies and thank you for the detailed explanation.

I'll try to think of an efficient way to fix the cosmetic issues you pointed out in a later PR. I don't want to rush it with a messy fix.

Generally it shouldn't be difficult at all to handle them, especially that we can now track it more accurately with the `playing_and_seeking` state.

Please don't hesitate to mention any other issues if found. I'll merge the PR sometime tomorrow, nevertheless, there is no rush in testing.

Keith94:
Okay sounds good to me :D

>Please don't hesitate to mention any other issues if found. I'll merge the PR sometime tomorrow, nevertheless, there is no rush in testing.

I only did a basic testing so far but it seemed fine. It affects just the seekbar properly now.

In your PR there's a small typo: "seedkbar"

You may want to mention in the docs that this new option only works properly when bottomhover is **disabled**.
For example, if bottomhover is turned ON and your mouse leaves the "bottom hover zone" the OSC (and tooltip/thumbnail) will hide anyway, even if the force seek tooltip option is set to yes.

Samillion:
Oops, nice catch. Thank you.

Keith94:
No worries

On second thought, it technically does work with bottom hover on or off, but it's a lot more _effective_ with bottom hover off, since your mouse can drift anywhere and still show that info.

I suppose the comment is not needed after all (sorry) haha but up to you :)

PR #339, opened 2025-02-08
refactor: don't register a shutdown handler

guidocella:
refactor: strip trailing whitespace

refactor: don't register a shutdown handler

shutdown is sent when mpv quits, but also to builtin script unloaded at
runtime with e.g. set osc no. There is no need to unset properties at
shutdown for external scripts since they only receive it on quit.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -3291,10 +3291,6 @@ tick = function()
     end
 end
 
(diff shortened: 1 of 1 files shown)


Samillion:
Beautiful, thank you very much. <3

PR #373, opened 2025-04-06
feat: add osc transparency options

Hellboy00000:
Add two new options: 
- `fade_transparency_strength`
- `window_fade_transparency_strength`

This adds more customization to the fade options and can be used to create a transparency box for the top and bottom of the OSC.

Edit: Config settings example

**Botton box**
```EditorConfig
# alpha of the OSC background box 
# This sets the size of the box. The higher the number, the larger the box
fade_alpha=80
# blur strength for the OSC alpha fade. caution: high values can take a lot of CPU time to render
# Set this to 0 to create the box
fade_blur_strength=0
# use with "fade_blur_strength=0" to create a transparency box
# This sets the amount of transparency
fade_transparency_strength=75
```

**Top box**
```EditorConfig
# alpha of the window title bar (0 to disable)
# This sets the size of the box. The higher the number, the larger the box
window_fade_alpha=152
# blur strength for the window title bar. caution: high values can take a lot of CPU time to render
# Set this to 0 to create the box
window_fade_blur_strength=0
# use with "window_fade_blur_strength=0" to create a transparency box
# This sets the amount of transparency
window_fade_transparency_strength=75
```

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -119,35 +119,37 @@ Create `modernz.conf` in your mpv script-opts directory:
 
 ### Colors and style
 
-| Option                      | Value     | Description                                                                                       |
-| --------------------------- | --------- | ------------------------------------------------------------------------------------------------- |
-| osc_color                   | `#000000` | accent color of the OSC and title bar                                                             |
-| window_title_color          | `#FFFFFF` | color of the title in borderless/fullscreen mode                                                  |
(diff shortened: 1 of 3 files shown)


Samillion:
Great work, thank you very much.

If I may, in the PR post, could you include an example config on creating a box. So that others can refer to the PR for reference in the future.

Hellboy00000:
@Samillion 

I added an example of the configuration with a small explanation.

Samillion:
Much appreciated.

Issue #374, opened 2025-04-07
Add a new Menu button


xiaoleichen:
### Expected behavior of the wanted feature

The built-in osc added a Menu button (new in v0.40?) that brings up a list of common action items.

<img width="146" alt="Image" src="https://github.com/user-attachments/assets/23f20b2b-0d14-4cee-9d17-269ce962fc2a" />

![Image](https://github.com/user-attachments/assets/0f83dc1f-e4f5-4ae5-9563-352fdc144c09)

Samillion:
Should we make the right click on playlist button show the new menu by default? Currently it just shows a text representation of the playlist.

Samillion:
In all honesty, I fully agree with you. In fact, I prepared for this a couple of months ago and there is even a dedicated icon for the new select menu in ModernZ:
https://github.com/Samillion/ModernZ/blob/b3c56c8d32fb1603df7603df0035dd68a73d8e4d/modernz.lua#L289

My thoughts at the time:
- Remove audio/sub buttons (or at least hide them by default)
- Keep playlist button
- Add the menu button (a vertical 3 dots icon) left of the playlist button

It would de-clutter the osc, introduce more options as I know @guidocella is considering adding even more options to the menu, and even if not, it still holds much more.

The only reason I didn't implement it is that I'm genuinely unsure if that's my personal preference/bias which users will dislike or not.

guidocella:
Yeah it should at least receive the new items already implemented in https://github.com/mpv-player/mpv/pull/15845 and https://github.com/mpv-player/mpv/pull/16129

Samillion:
The good thing is: subtitle/audio selection in many online players are now within a menu and not in dedicated buttons, so users wouldn't have a UX shock.

The hesitation is purely from the standard within mpv oscs ecosystem. I'd say majority of available OSCs have dedicated buttons for audio/sub, but that shouldn't dictate it as a rule, I guess.

Guido's menu offers so much more in terms of control, and my god with that file dialogue and shortcuts, it's a major QoL addition.

I'll leave this open for discussion, if no objection(s) by next Monday when I come back home, I'll implement the new changes.

Hellboy00000:
If possible don't remove audio/sub buttons, better hide them by default.

Anthonyy232:
Agree with hiding them by default, but would love to be able to enable them again. I use the audio/video buttons a TON -- in fact they are pretty much the only buttons I use aside from the information button.

Issue #368, opened 2025-04-02
Transition to mpv 0.40


lostb053:
### Expected behavior of the wanted feature

I don't know if you already planned this, but new mpv now comes with bunch of in-built menu and UI based access to certain aspects of player like subs, audio, and a better context menu.

Modernz's menu earlier was a great option earlier when we had nothing, but would be nice to see the changes based on new implementation.

Keith94:
It doesn't have a dedicated button for it right now, but it's functional if you add some key bindings

```
MBTN_LEFT  cycle pause
MBTN_RIGHT script-binding select/menu; script-message-to modernz osc-hide
```

i.e. left-click to pause, right-click for context menu

Keith94:
No scripting is needed. Just need to create an input.conf file and paste the two lines above.

See https://www.ghacks.net/2019/07/27/how-to-customize-the-settings-in-mpv/

PR #375, opened 2025-04-17
refactor: button positioning for easier addition and shuffling of buttons

xiaoleichen:
Dynamically compute positions and visibility of left and right buttons (for video only).
This should make adding and moving around buttons easier.

### DIFF
--- a/extras/locale/modernz-locale.json
+++ b/extras/locale/modernz-locale.json
@@ -25,6 +25,7 @@
         "download_in_progress": "Téléchargement en cours",
         "downloading": "Téléchargement",
         "downloaded": "Déjà téléchargé",
+        "menu": "Menu",
     },
     "de": {
         "idle": "Dateien oder URLs hier ablegen, um sie abzuspielen",
@@ -52,6 +53,7 @@
         "download_in_progress": "Herunterladen läuft",
         "downloading": "Herunterladen",
         "downloaded": "Bereits heruntergeladen",
+        "menu": "Menü",
     },
     "es": {
         "idle": "Suelta archivos o URLs aquí para reproducirlos",
@@ -79,6 +81,7 @@
         "download_in_progress": "Descarga en curso",
         "downloading": "Descargando",
         "downloaded": "Ya descargado",
+        "menu": "Menú",
     },
     "pl": {
         "idle": "Upuść plik lub łącze URL do odtworzenia",
@@ -106,6 +109,7 @@
         "download_in_progress": "Pobieranie w toku",
         "downloading": "Pobieranie",
         "downloaded": "Już pobrano",
+        "menu": "Menu",
     },
     "jp": {
         "idle": "ファイルやURLのリンクをここにドロップすると再生されます",
@@ -133,6 +137,7 @@
         "download_in_progress": "ダウンロード中",
         "downloading": "ダウンロード中",
         "downloaded": "既にダウンロード済み",
+        "menu": "メニュー",
     },
     "zh": {
         "idle": "将文件或URL放在这里播放",
@@ -188,6 +193,7 @@
         "download_in_progress": "التنزيل جارٍ",
         "downloading": "جاري التنزيل",
         "downloaded": "تم التنزيل بالفعل",
+        "menu": "قائمة",
     },
     "ru": {
         "idle": "Перенести файл или URL для воспроизведения",
@@ -215,5 +221,6 @@
         "download_in_progress": "Загрузка в процессе",
         "downloading": "Загрузка",
         "downloaded": "Уже загружено",
+        "menu": "Меню",
     },
 }
--- a/extras/locale/modernz-locale.json
+++ b/extras/locale/modernz-locale.json
@@ -160,6 +160,7 @@
         "download_in_progress": "下载中",
         "downloading": "正在下载",
         "downloaded": "已下载",
+        "menu": "菜单",
     },
     "ar": {
         "idle": "أسقط الملفات أو الروابط هنا للتشغيل",
--- a/modernz.conf
+++ b/modernz.conf
@@ -125,9 +125,9 @@ volume_control_type=linear
 # show playlist button: Left-click for simple playlist, Right-click for interactive playlist
 playlist_button=yes
 # hide playlist button when no playlist exists
-hide_empty_playlist_button=yes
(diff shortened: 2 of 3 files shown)


xiaoleichen:
I think we could possibly permanently repurpose the hamburger playlist button as the general menu button, because right click of the playlist previous / next buttons already can bring up the playlist.

This behavior is consistent with the new stock OSC.

Samillion:
Excellent work, thank you very much.

> Even if we don't do that, I'd like to see the playlist button as the first button on the left. Logically, we have file level selection playlist button at the left most position and then audio, subtitle level selection buttons etc.

As I mentioned in the issue, I genuinely agree with you, my only concern is user reaction to this change. Do you mind if we transition gradually?

For example, you could change the default right click on playlist button to display the menu:
```EditorConfig
playlist_mbtn_right_command=script-binding select/menu; script-message-to modernz osc-hide
```
If you agree, then please go ahead and commit the change and then I will merge.

The only reason I say this, `console/select` has been around for a long while now. The creator (@guidocella) and I had to tell so many users that it exists and sometimes we had to defend it over the plain text playlist (I still can't believe we had to defend it lol)

BUT, I will not shy away from progress for a fear of weak arguments.

I believe 100% this will be an improvement, and the new select menu should be included, whether as a dedicated button or a change to playlist button behavior.

In essence, I just want a gradual transition to it, as I know how mpv users can be fussy about changes.

Samillion:
I'm overthinking this, I mean, it makes sense, why delay progress.

Please go ahead make the commit and change the left click for playlist as menu, and the right click as playlist menu.

xiaoleichen:
Committed the change. But It might still need a bit cleanup. Now `hide_empty_playlist_button` and `gray_empty_playlist_button` are disabled by default to always show the button. We could just remove those options. Also, the tooltip still says "playlist". It probably should be something like "Menu"

xiaoleichen:
Take your time. 

By "cleanup", I meant removing those two hide/gray playlist options and updating the tooltip to "Menu". 

We can keep those two options if you prefer. I'm putting up an additional commit to update the tool tip. Then it should be all good to merge from my side.

Samillion (approved):

Samillion:
Thank you again. Merging.

novaexe:
@Samillion while i mostly agree the hamburg menu should always be visible and at the farthest left, the menu itself should remain as a playlist until more features get added, this is because always having to right click for a playlist option in three separate places just feels weird and no offense but this barely counts as a "menu" 
image
when things such as https://github.com/tsl0922/mpv-menu-plugin exist
so my suggestion is simple: to swap right click and left click funts, thankfully you didn't get rid of the hide/grey for people that just want it as a playlist.
again don't get me wrong this is overall a good change but currently the ease of use should take priority

Samillion:
> so my suggestion is simple: to swap right click and left click funts

This is already possible, simply add the following to your `modernz.conf`
```EditorConfig
playlist_mbtn_left_command=script-binding select/select-playlist; script-message-to modernz osc-hide
playlist_mbtn_right_command=script-binding select/menu; script-message-to modernz osc-hide
```

If we get complaints, we can simply switch to it as the default behavior in the future.

Thankfully though, the way everything was set up when we started adding buttons is that each button can do whatever function the user wants.

> no offense but this barely counts as a "menu"

It's definitely in the early stages. A couple of months ago it didn't even exist. Guido has been working gradually, and a few things might be added to it soon (including open dialogue)

novaexe:
wow ur fast on the reply :D, yea i think the menu has potential especially if you could somehow make it appear on the far left side instead of dead center of the screen above the seekbar title for example 
image

Samillion:
> if you could somehow make it appear on the far left side instead of dead center of the screen above the seekbar title for example

I remember discussing this with @guidocella on irc. If I remember correctly, uosc was the inspiration for the new look and center positioning.

He experimented with multiple locations, and even centered text. Maybe a position option, I'm not sure. I added position options to the mute indicator in https://github.com/Samillion/ModernZ/pull/383 which is essentially a list of `ass tag an` values

```lua
    local mute_pos_list = {
        ["top_left"] = 7,
        ["top_center"] = 8,
        ["top_right"] = 9,
        ["middle_left"] = 4,
        ["middle_center"] = 5,
        ["middle_right"] = 6,
        ["bottom_left"] = 1,
        ["bottom_center"] = 2,
        ["bottom_right"] = 3,
    }
    local mute_pos = mute_pos_list[options.mute_indicator_pos:lower()] or 6
```

I have high hopes, to be honest. The work on console and select has been impressive and a game changer. In time I am sure it'll grow even further.

> wow ur fast on the reply :D

Tehehe

Samillion:
Nevertheless, please feel free to share any function changes and opinions. None of this is set in stone, I am fine with changing the default behavior of any option if the users ask for it.

It's just hard sometimes to decide on default behavior, since some will like it, others won't. For example the menu being main was requested in https://github.com/Samillion/ModernZ/issues/374 and https://github.com/Samillion/ModernZ/issues/368

The only thing that makes me not panic, is that users have full control to change it within the config, easily.

novaexe:
"It's just hard sometimes to decide on default behavior, since some will like it, others won't." yep cant please everyone xD, while i personally dont want a hamberg menu when i have the mpv-menu plugin i understand not everyone does or can since it's kinda a windows only thing, my only real concern is im not sure the menu can be anything like (fully featured as) the plugin or mpv.net's amazing config editor https://github.com/mpvnet-player/mpv.net/raw/main/docs/img/ConfEditor.webp

novaexe:
kinda where uosc failed a little bit imo, feels very clunky to use when diving into sub-menu's

novaexe:
but then if i could figure out how to do it i would have right click previous as open reciently played, right click on next would bring up open new playlist folder, and the "menu" button would just bring up current playlist on left click and a menu with video manipulation on right click xD

novaexe:
that would bring far more features to the next and prev than just a triple redundant feature xD

guidocella:
I already implemented a real context menu

screenshot

Just waiting for https://github.com/mpv-player/mpv/pull/16129 to be merged

Samillion:
Incredible, I knew my high hopes were justified.

If/when guido implements fluent icons (or any new ones) he showcased a while ago, the stock osc experience will be perfect.

PR #391, opened 2025-06-03
feat: option to hide osc while cursor is in bounds

### LINKED ISSUE #382: Let the OSC hide even while being hovered

Currently If mouse cursor is hovering the top or bottom geometry, the OSC stays visible.

My idea is to let the OSC hide anyway if the mouse cursor is resting in the osc boundaries, so the user does not have to pay close attention to the last position it was at.

This would be based on hidetimeout value.

Keith94:
Fixes #382


Samillion:
Thank you very much for your efforts.

PR #404, opened 2025-10-21
locale: Add Icelandic locale

finnurthorisson:
As title says


Samillion:
Awesome, thank you very much.

PR #412, opened 2025-12-27
fix: 5s jump icon for Material theme

mhyee:
Previously, the wrong code points were being referenced.

Tested with the following `modernz.conf`:
```
icon_theme=material
jump_amount=5
```

Before:
image

After:
image


Samillion:
Oh, wow. Great catch. Thank you very much.

PR #434, opened 2026-02-01
fix: zoom descriptions

Keith94:


Samillion:
Oh, wow. Nice catch. Thank you very much. ❤️

PR #435, opened 2026-02-01
feat: add modern-compact layout

### LINKED ISSUE #425: Better default layout

Functions related to playback and seeking to go at the start.
All extra buttons, i.e audio, subtitle, volume, etc. grouped at the end.

For example:

Image

nekoxuee:
Add a new alternative layout called modern-compact.

closes #425

preview

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -15,6 +15,7 @@ Create `modernz.conf` in your mpv script-opts directory:
 | Option         | Value           | Description                                                                                                                    |
 | -------------- | --------------- | ------------------------------------------------------------------------------------------------------------------------------ |
 | language       | en              | set language (for available options, see: [Translations](https://github.com/Samillion/ModernZ/blob/main/docs/TRANSLATIONS.md)) |
+| layout         | modern          | set layout: `"modern"` or `"modern-compact"`                                                                                   |
 | icon_theme     | fluent          | set icon theme. accepts `fluent` or `material`                                                                                 |
 | font           | mpv-osd-symbols | font for the OSC (default: mpv-osd-symbols or the one set in mpv.conf)                                                         |
 | idlescreen     | yes             | show mpv logo when idle                                                                                                        |
--- a/modernz.conf
+++ b/modernz.conf
@@ -1,6 +1,8 @@
 # Language and display
 # set language (for available options, see: https://github.com/Samillion/ModernZ/blob/main/docs/TRANSLATIONS.md)
 language=en
+# set layout: "modern" or "modern-compact"
+layout=modern
 # set icon theme. accepts fluent or material
 icon_theme=fluent
 # font for the OSC (default: mpv-osd-symbols or the one set in mpv.conf)


Samillion:
That is a beautiful compact layout, seriously well done.

Seekbar position is a bit different for me than shown in your preview, not sure why. Is it an issue on my end?

| `refY - 82` (current)  |
| ------------ |
| image |

| `refY - 72` (adjusted)  |
| ------------ |
| image |

Samillion:
Also, as I'm sure he'll love to see this, I have to CC @Keith94

nekoxuee:
> That is a beautiful compact layout, seriously well done.
> 
> Seekbar position is a bit different for me than shown in your preview, not sure why.
> 
> `refY - 82`:
> 
> image
> and this is with `refY - 72`:
> 
> image
> Is it an issue on my end?

Yeah, that was my mistake. It was supposed to be `refY - 72`, I was testing changes across two different files in my config folder and the cloned repo folder and might have undone something while editing the files.

Samillion:
Ah, I've done that many times 😅. I genuinely love this layout, pretty sure it'll be my default.

Let me know whenever it's ready to merge. Take all the time you need.

nekoxuee:
okay, now should be fine now. I always find it a bit of a headache to rebase to keep the commits tidy

Keith94:
Thanks for adding that so fast! Much nicer overall.

I did a quick check and noticed a few things compared to the modern layout.

* The starting and ending position of the buttons is more "pushed in" towards the middle.
* Volume bar is noticeably thinner (vertically).
* Chapter title is always shown even if turned off.
* Audio and subtitle positions are swapped around.
* Chapter skip, jump, info buttons are always hidden. Maybe that's expected, or can be they visible on the right side?

Samillion:
> * Chapter skip, jump, info buttons are always hidden. Maybe that's expected, or can be they visible on the right side?

Expected, definitely. As this is meant to be compact. Menu and context menu give access to all the extra stuff.

We can always adjust any additions in the future if needed. Consider this a base start for the layout.

nekoxuee:
> * The starting and ending position of the buttons is more "pushed in" towards the middle.
> * Volume bar is noticeably thinner (vertically).

Both of these are used in other player ui's because of visual hierarchy. I can make them like they were before if that is preferred.
<br>
> * Chapter title is always shown even if turned off.

I completely forgot there was a user option for that. Will add it now.
<br>
> * Audio and subtitle positions are swapped around.

I can also change that.
<br>
> * Chapter skip, jump, info buttons are always hidden. Maybe that's expected, or can be they visible on the right side?

Those are not even added since the idea was to use the menu button (or hotkeys) in case you want to access the less frequent actions

Samillion:
> Those are not even added since the idea was to use the menu button (or hotkeys) in case you want to access the less frequent actions

Indeed. I think in the future the most requested will be chapter skip and jump buttons. Anything else is as conveniently accessed through menu/context menu.

Nevertheless, it's definitely not urgent.

Keith94:
>The starting and ending position of the buttons is more "pushed in" towards the middle.
Volume bar is noticeably thinner (vertically).
Both of these are used in other player ui's because of visual hierarchy. I can make them like they were before if that is preferred.

I see what you mean, and it's started to grow on me. These new values could be applied to the other modern and modern-image layouts too, no? At some point anyway.

>Those are not even added since the idea was to use the menu button (or hotkeys) in case you want to access the less frequent actions

This is OK since I like a cleaner look anyway. Might be worth explaining the differences (ie button omissions) between the layouts somewhere in the docs etc

Samillion:
> This is OK since I like a cleaner look anyway. Might be worth explaining the differences (ie button omissions) between the layouts somewhere in the docs etc

Agreed. I'll definitely make a big section about this layout and its details in the next release as well.

Samillion:
> I see what you mean, and it's started to grow on me. These new values could be applied to the other modern and modern-image layouts too, no? At some point anyway.

The volume bar is too thin, I agree to that. The starting position can be whatever is decided to be honest. You guys have a much better grasp on UI hierarchy than I do.

Keith94:
Indented start_x is good by me IMO.
I also feel that the volume bar width should be reduced to match something like YouTube's. It takes up a lot of real estate especially for a compact mode.

nekoxuee:
Made the volume bar shorter and thicker, swapped the position of the audio and subtitle buttons, and fixed the chapter title not hiding when the user option was set to false

Keith94:
Thanks for the fixes. The script should probably check whether user toggled these ones off, too

playlist_button=no
subtitles_button=no
ontop_button=no
fullscreen_button=no

Keith94:
This is not urgent either, but supporting vertical videos would need some tweaking in the future

nekoxuee:
> This is not urgent either, but supporting vertical videos would need some tweaking in the future

I'll need help with this because I don't know the correct way to do it

Samillion:
Thanks again for all the effort. I will merge, and start working on any additions//adjustments in the next couple of days.

PR #445, opened 2026-02-14
refactor: simplify keep with cursor

Keith94:
Code is a bit simpler and still works.


Samillion:

Issue #460, opened 2026-02-22
osc crash with 'attempt to perform arithmetic on field showtime'


Keith94:
### mpv Information

```bash
mpv v0.41.0-dev-g4157fc682 Copyright © 2000-2026 mpv/MPlayer/mplayer2 projects
 built on Feb 19 2026 16:11:16
libplacebo version: v7.360.0
FFmpeg version: 1834eb4
FFmpeg library versions:
   libavcodec      62.11.100
   libavdevice     62.1.100
   libavfilter     11.4.100
   libavformat     62.3.100
   libavutil       60.8.100
   libswresample   6.1.100
   libswscale      9.1.100
```

### Reproduction Steps

Set the option `seek_resets_hidetimeout=false`

### Expected Behavior

No crash

### Actual Behavior

<img width="2347" height="1157" alt="Image" src="https://github.com/user-attachments/assets/2c0fa142-311c-4599-a072-76514a743c5a" />

### Log File

[output.txt](https://github.com/user-attachments/files/25463393/output.txt)

### I carefully read all instructions and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information.

Samillion:
Nice catch. This is caused by `state.showtime` being `nil` at initiation, I assume because it's not valued fast enough sometimes at mpv start. 

I'm not at home, could you try changing this:
```lua
local state = {
    showtime = nil,                         -- time of last invocation (last mouse move)
```

To this:
```lua
local state = {
    showtime =  mp.get_time(),              -- time of last invocation (last mouse move)
```

Or this:
```lua
local state = {
    showtime =  0,                          -- time of last invocation (last mouse move)
```

Samillion:
Hmmm. Would it be better to adjust `function process_event(source, what)` instead? Since that's where the crash is triggered.

For example, changing this:
```lua
                        if get_hidetimeout() >= 0 and get_touchtimeout() <= 0 then
                            local elapsed_time = mp.get_time() - state.showtime
```

To this:
```lua
                        if get_hidetimeout() >= 0 and get_touchtimeout() <= 0 then
                            local elapsed_time = mp.get_time() - (state.showtime or mp.get_time())
```

Samillion:
Hmmmmm. OR (lol, sorry) in `function process_event(source, what)` change this:
```lua
                        if get_hidetimeout() >= 0 and get_touchtimeout() <= 0 then
[...]

PR #465, opened 2026-02-24
feat: independent show/hide for window controls

### LINKED ISSUE #317: Show only OSC or title bar when hovering

In borderless/fullscreen mode, allow hovering on OSC to show only OSC and not the title bar, and vice versa.

IAmHal9000:
Adds a new independent_wc option that lets the OSC and window controls bar show and hide independently. When enabled, the bottom hover zone triggers the OSC and the top hover zone triggers the window controls, each with their own fade and autohide timers.                                                                                 
                                                                                                                                                                          
  New options:                                                                                                                                                            
  - independent_wc (default: no) - enable independent show/hide for window controls
  - tophover_zone (default: 40) - height of the top hover zone in pixels, used when independent_wc is enabled

  Extended option:
  - osc_on_start now accepts no, bottom, top, or both to control which zones are shown on file load. Boolean yes/no values are still accepted for backwards compatibility.

  When independent_wc is disabled (the default), behavior is identical to before - nothing changes for existing users.
  
  Closes #317

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -4270,6 +4270,11 @@ local function validate_user_opts()
             table.insert(state.visibility_modes, str)
         end
     end
+
+    if user_opts.independent_wc and not user_opts.bottomhover then
+        msg.warn("independent_wc requires bottomhover. Setting bottomhover=yes.")
+        user_opts.bottomhover = true
+    end
 end
 
 -- read options from config and command-line


Samillion:
Just did a quick test. Great work, looks great and works great.

I found a little issue. Add the following to `modernz.conf`
```EditorConfig
independent_wc=yes
showonpause=yes
keeponpause=yes
```

Pause then unpause, Window Controls (top bar) hides immediately while OSC (bottom bar) respects `hidetimeout`.

We had a similar issue which was fixed by https://github.com/Samillion/ModernZ/pull/308

IAmHal9000:
Done. 

The fix is in. The root cause was that always_on(false) is a no-op when state.enabled = false (which is set at the top of the pause handler before visibility_mode is called), so neither OSC nor WC are hidden on unpause - they just keep their old showtimes from when the pause started. OSC survives because the mouse is hovering   the bottom, WC doesn't because nothing is hovering the top. Resetting wc_showtime gives it the same fresh window OSC gets implicitly from mouse hover protection.

Keith94:
The fadein feature is not working anymore in this build, even when set to "true" it acts like "false".

```lua
fadein = true, -- whether to enable fade-in effect
```

IAmHal9000:
I have pushed a fix and tested some more. please test and review

Samillion:
## Issue
```EditorConfig
independent_wc=yes
showonpause=yes
keeponpause=yes
```

Pause, wait 2 seconds (more than hidetimeout), then unpause.

OSC hides faster than WC.

## Cause
```lua
            -- reset wc timer so it gets a fresh hidetimeout on unpause,
            -- matching the effective behaviour of the osc which is protected
            -- by mouse hover when the cursor is over the bottom zone
            if user_opts.independent_wc and state.wc_visible then
                state.wc_showtime = mp.get_time()
            end
```

## Solution
Should add:
```lua
if state.osc_visible then state.showtime = mp.get_time() end
```

To make the OSC timeout match WC. It's better to test this without hovering either zone, as in mouse out of window. To see `hidetimeout` behavior.

Since this activates by `visibility_mode` set to `always` on pause, hover wouldn't affect it. Then comment should be:
```lua
            -- reset timer so it gets a fresh hidetimeout on unpause
```

End result:
```lua
            -- reset timer so it gets a fresh hidetimeout on unpause
            if state.osc_visible then state.showtime = mp.get_time() end
            if user_opts.independent_wc and state.wc_visible then
                state.wc_showtime = mp.get_time()
            end
```

What do you think?

IAmHal9000:
I agree with you and I feel a bit dumb for not catching it myself

Samillion:
> I agree with you and I feel a bit dumb for not catching it myself

Not at all. This is a big feature update, small hiccups are expected. In fact, it is honestly impressive that so far only those two issues were found. Great work. ❤️ 

@IAmHal9000 @Keith94 Do you guys mind if I keep this PR opened longer than usual? So we can all test this more thoroughly in the next couple of days.

Since it is a big update, I'd like to ensure a smooth experience, as possible.

I'll help wherever I can.

IAmHal9000:
@Samillion fine by me

Keith94:
Sometimes I see a weird desync issue, If you use `osc_on_start=both`, the WC hides slightly faster than the OSC, even though they should use the same hidetimeout.

https://github.com/user-attachments/assets/b117c6ae-3a50-48ef-9257-9e944794ecd4

Samillion:
> Sometimes I see a weird desync issue, If you use `osc_on_start=both`, the WC hides slightly faster than the OSC, even though they should use the same hidetimeout.

In `local state`, change
```lua
wc_showtime = nil
```

To:
```lua
wc_showtime = mp.get_time()
```

Fixes it?

This is probably related to https://github.com/Samillion/ModernZ/pull/461 when we initiated `showtime` as a time value instead of `nil` to avoid crash. Which would explain the difference with WC timing with `osc_on_start`... possibly. I haven't tested, as I'm not home right now.

Now I'm wondering if there is a better fix than that PR... hmmmmmm

IAmHal9000:
I am mirror the fix from #461 to wc_showtime

Samillion:
Thank you. Which hopefully is going to be a temporary change that I'll fix later. I re-opened that issue and suggested a more rational fix

https://github.com/Samillion/ModernZ/issues/460#issuecomment-3963858436

Keith94:
Follow up to https://github.com/Samillion/ModernZ/pull/465#issuecomment-3963564780 the desync only happens with `seek_resets_hidetimeout = true`. Set to false and issue is fixed.

I've just tested this with commit https://github.com/Samillion/ModernZ/pull/465/changes/8af708a4548d09382798f7f4d8845a867af6d33e where the `wc_showtime = nil,` is still present.

Samillion:
Seems it happens when `seek_resets_hidetimeout=yes`

```lua
mp.observe_property("seeking", "native", function(_, seeking)
    if user_opts.seek_resets_hidetimeout then
        reset_timeout()
    end
end)
```

## Solution
This:
```lua
local function reset_timeout()
    state.showtime = mp.get_time()
end
```

Should be:
```lua
local function reset_timeout()
    state.showtime = mp.get_time()
    state.wc_showtime = mp.get_time()
end
```

Samillion:
By the way, @Keith94, this is why I call you detective. In a million freaking years, I would have never found this issue lol.

IAmHal9000:
updated the PR

modernz.lua
```diff
@@ -564,7 +564,7 @@ local state = {
     anitype = nil,                          -- current type of animation
     animation = nil,                        -- current animation alpha
     wc_visible = false,                     -- window controls visibility (independent mode)
-    wc_showtime = nil,
+    wc_showtime = mp.get_time(),
```
Keith94: Don't you have to revert the wc_showtime state back to nil since Sam is using a different approach for it.

Samillion:
Awesome work so far. I'll do heavy testing tomorrow, but so far everything's genuinely looking promising. Well done.

Keith94:
Whenever I click on a button or drag the seekbar, it resets the hidetimeout for the WC. For example, with WC visible, click and hold the jump forward button. The WC never hides, but maybe it should.

Samillion:
> Whenever I click on a button or drag the seekbar, it resets the hidetimeout for the WC. For example, with WC visible, click and hold the jump forward button. The WC never hides, but maybe it should.

Hmmm. That's a thinker. Should it hide on its own? If so, what would be the parameters for `reset_timeout()`?

If `reset_timeout()` shouldn't reset WC timer at all, then something like this?

```diff
local function reset_timeout()
    state.showtime = mp.get_time()
-   state.wc_showtime = mp.get_time()
end
```

```diff
mp.observe_property("seeking", "native", function(_, seeking)
    if user_opts.seek_resets_hidetimeout then
        reset_timeout()
+       state.wc_showtime = mp.get_time()
    end
end)
```

**Psuodo logic:**
- WC hide timeout will only be reset with `seek_resets_hidetimeout=yes` on seek actions. Any other reset timeout calls will only reset OSC timer.
- WC timer will be tied to its last visible state (ie: on hover, on pause (with `keeponpause` or `showonpause`)...etc)

Is that a better behavior?

Samillion:
Actually, should WC timer be reset with `seek_resets_hidetimeout=yes`? What benefit does keeping WC visible with seek?

Samillion:
Sorry for the sequence of comments. 

Realistically speaking, WC (when independent hover is on) should only be visible on hover, no?

So the following scenarios should only affect OSC:
- Start actions (ie: osc on start)
- Seek actions
- Show/keep on pause actions
- Visibility mode: `always` (not sure about this one)

Unless there is a use case I'm not thinking of for showing WC in those other scenarios.

To be fair though, all these are behavior questions, which I can adjust in the future as we see fit. Not necessarily in this PR I mean.

This PR should focus on any bugs or crashes, as in a stable function of independent show/hide.

Keith94:
> So the following scenarios should only affect OSC:
> 
> * Start actions (ie: osc on start)
> * Seek actions
> * Show/keep on pause actions
> * Visibility mode: `always` (not sure about this one)

I like your logic; it should mostly stay out of the way and only show up when we request it.

The only time I care for it to trigger by itself is on file start. My reasoning is binge watching a series on auto play, it's nice to see the title/filename of a new episode when it flashes on screen. If I'm not mistaken this could be achieved with OSD as well. I'll have to check.

Can address it after so this doesn't get delayed.

Samillion:
> if I'm not mistaken this could be achieved with OSD as well. I'll have to check.

I use: (`mpv.conf`)
```EditorConfig
osd-playing-msg="${!playlist-count==1:[${playlist-pos-1}/${playlist-count}] }${?demuxer-via-network==yes:${media-title}}${?demuxer-via-network==no:${filename/no-ext}}"
```

Keith94:
> > if I'm not mistaken this could be achieved with OSD as well. I'll have to check.
> 
> I use: (`mpv.conf`)
> 
> ```editorconfig
> osd-playing-msg="${!playlist-count==1:[${playlist-pos-1}/${playlist-count}] }${?demuxer-via-network==yes:${media-title}}${?demuxer-via-network==no:${filename/no-ext}}"
> ```

That's a really good alternative. :) Thanks!

Samillion:
@IAmHal9000 @Keith94 I know I'm hesistant to merge this quickly, but honestly I can't find any breaking issues.

Have any of you found anything that needs to be fixed or adjusted?

For behavior stuff, we can add them one by one later on as we go.

Keith94:
>Have any of you found anything that needs to be fixed or adjusted?

It's very broken if you set your options like this.

`bottomhover = false`
`independent_wc = true`

When hovering over OSC, the WC shows up and hides on its own as per hidetimeout.
Any further mouse movement makes the WC instantly show up again.

The reverse is also true when hovering the WC instead.

Samillion:
Hmmm. This is a thinker. The problem with this is that `bottomhover=no` makes it so that any mouse movement, anywhere, shows the OSC.

In essence, making `independent_wc` redundant. The name here makes it confusing. 

If we name it based on actual behavior, `bottomhover` is `anyhover_show_osc`.

What I can't determine is, what do we do here? My initial thoughts is that we add in validating options the following:
- If `bottomhover=false`, set `independant_wc=false`
- If `independant_wc=true`, set `bottomhover=true`

But, I'm not sure if that actually makes sense or just my brain simplifying it.

Keith94:
Sounds like that should avoid the conflict that I encountered.

I would personally be happy with just the two behavior options (proximity-based hover and "show everything on each mouse move" hover). The old bottomhover behavior feels a little redundant now, but removing it could upset some people, I guess?

IAmHal9000:
Found the conflict in the code. When bottomhover=false the entire window becomes a single hover zone, so every mouse move triggers both show_osc() and show_wc(), making independence pointless.                                                                                                                                                
                  
Simplest fix is option validation. If independent_wc=yes and bottomhover=no, force bottomhover=yes with a warning. Same pattern already used elsewhere in validate_user_opts().

Samillion:
Sounds good to me.

IAmHal9000:
I have pushed a fix

Keith94:
I can't break it anymore after that. Everything appears solid to me.

Samillion:
I'll merge on Monday to give us all a bit more time to daily-drive this for any possible scenarios.

Wonderful work so far.

Samillion:
Merging. Thank you very much for your effort.

We can adjust anything as we go.

PR #469, opened 2026-02-27
feat: add an option to hide OSC with select open

Keith94:
Observes the property `user-data/mpv/console/open` to show/hide OSC instead of using script messages.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -3878,6 +3878,7 @@ end)
 mp.observe_property("user-data/mpv/console/open", "bool", function(_, val)
     if val and user_opts.visibility == "auto" and not user_opts.showonselect then
         osc_visible(false)
+        wc_visible(false)
     end
 end)
 mp.observe_property("display-fps", "number", set_tick_delay)


Samillion:
Nicely done. However, this would break function in most Linux mpv releases, as this is a fairly recent PR at mpv.

The one thing Linux users suffer is having an outdated mpv package due to distro policies. Surprising that Windows is actually the best at this one.

Keep the PR open, please. I'll try to think of a solution. For example, we can wrap it with a property check around that observer (if open exists then observe). But I can't think of a solution to the default button(s) function values, since they just no longer have the hide option.

Keith94:
I didn't realize that, but it's understandable. The way you have it now is a good workaround that reaches a large audience. I'm okay with either method.

Samillion:
I'm stupid. This wouldn't break anything. I thought it would lua crash due to property not existing in older builds, but I just realized this just checks `user-data/`. 

So even on older mpv, this would just silently not trigger hide from the observer. As in, it'll always remain in a `val = false` state. Which is fine.

What is worth testing here is removing `; script-message-to modernz osc-hide` from all select/console buttons where it's used, and not include the observer.

In essence, to see behavior when osc never hides with long select menus. To simulate older builds without the `open` observer being `true`.

I'll test this when I'm back home in a few hours.

Samillion:
It should work. The behavior (if I'm right), in older builds will be:
- On click: playlist/menu shows, osc hides as it should with `hidetimeout` as normal
- If `visibility` is set to `always` (ie: `keeponpause`), with a long playlist/menu, playlist/menu should render above osc
- Even on overlap, console has a higher layer grouping, so it's on top of osc/wc

Just putting this here as a reminder when I test.

Keith94:
The property was also added more than 1 year ago (13 months). https://github.com/mpv-player/mpv/commit/8669205d92d4362abc96aec68f797ab508961c27

Are Linux builds that far behind? I'm kinda surprised lol.

Samillion:
Woah. Why does it feel like it was last month.

> Are Linux builds that far behind? I'm kinda surprised lol.

Fedora and similar, 6 months, 12 months. Stable distros, 1 to 2 years.

https://github.com/Samillion/Windows-to-Linux

Samillion:
Either way, yeah, this is good to go. Mind if we merge this after the WC update?

That way we can include wc hide as well when it's up.

Keith94:
> Either way, yeah, this is good to go. Mind if we merge this after the WC update?
> 
> That way we can include wc hide as well when it's up.

Totally fine.

Samillion:
Awesome. Thank you very much.

Merging.

PR #481, opened 2026-03-03
refactor: reduce redundancy for hidetimeout

Samillion:
I am not sure what I was doing there at the time or why I repeated this code so many times needlessly.

**Changes**:
- Let the actual helpers handle hidetimeout instead of repeating it multiple times.

@Keith94 One example of repeated code and redundancy like I mentioned earlier. My initial tests shows everything working as it should.

But, knowing that my brain only pretends to exist, please let me know if I broke more things than fixed lol.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -3624,9 +3624,9 @@ local function process_event(source, what)
                         if in_top then show_wc() end
                     elseif in_bottom or in_top then
                         show_osc()
-                else
-                    state.touchtime = nil
-                end
+                    else
+                        state.touchtime = nil
+                    end
                 else
                     show_osc()
                     if user_opts.independent_wc then


Keith94:
Looks cleaner! I will definitely let you know, since there's a few(?) different use cases to check.

nekoxuee:
Hi, I'm catching up on my notifications now, sorry for missing the other ping.

I tried the script with different options and looks good to me.

I think it's a bit hard to notice redundancy because the script is a +4000 line monolith. I don't know if this has been talked about before, but I feel it would benefit from being split into components. Idk if that's common practice for lua scripts.

Samillion:
> I think it's a bit hard to notice redundancy because the script is a +4000 line monolith. I don't know if this has been talked about before, but I feel it would benefit from being split into components. Idk if that's common practice for lua scripts.

Indeed. It's been suggested before by commander as well in the past.

I found a few more redundant code, even unused code. After the cleanup we'll see the result.

It is definitely all over the place right now. I'm doing it in small PRs to not lose track.

> sorry for missing the other ping

No worries at all. I hope all is well with you. ❤️

Keith94:
Yeah, everything works fine for me too.

PR #482, opened 2026-03-04
fix: validate dynamic time for modern-compact and improve helper

Samillion:
**Changes**:
- Validate `dynamic` time option for modern-compact
- Improve (and move) `format_time()` to eliminate redundancy

/cc @Xurdejl You introduced this feature to ModernZ, I'd love your opinion on the changes, if it's not a bother.

Also an obligatory /cc to @Keith94 the detective I burden with all my troubles lol.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -477,7 +477,6 @@ if external then
     end
 end
 
-
 local icons
 local iconfont
 local function set_icon_theme()
@@ -1535,7 +1534,6 @@ local function render_elements(master_ass, osc_vis, wc_vis)
                     end
                 end
             end
-
         elseif element.type == "button" then
             local buttontext
             if type(element.content) == "function" then
@@ -3184,7 +3182,6 @@ local function osc_init()
         end
     end
     visible_min_width = visible_min_width + (user_opts.download_button and 100 or 0)
-
 
     -- cache info
     ne = new_element("cache_info", "button")


Keith94:
Functionality is still flawless like before.

The only (UI-related) issue with the time codes is the width is a bit small with really long time codes. Easy enough to PR later..

Samillion:
Awesome, thank you very much for testing. I'll merge in a few hours when I'm back home.

nekoxuee:
Looks good!!
I did something similar a few months ago when I updated my config. Initially just wanted to remove the leading zeros (e.g., 1:05 instead of 01:05), but ended up refactoring it to make it cleaner. Also moved it up so it's grouped with the other helper functions

``` lua
local function format_time_custom(seconds, with_ms)
    if seconds == nil then return "" end
    local h = math.floor(seconds / 3600)
    local m = math.floor((seconds % 3600) / 60)
    local s = math.floor(seconds % 60)

    local time_str
    if h > 0 then
        time_str = string.format("%d:%02d:%02d", h, m, s)
    else
        time_str = string.format("%d:%02d", m, s)
    end

    if with_ms then
        local ms = math.floor((seconds % 1) * 1000)
        time_str = time_str .. string.format(".%03d", ms)
    end

    return time_str
end
```

PR #483, opened 2026-03-04
fix: correct ASS tags, remove unused code, fix shuffle button

Samillion:
**Changes**:
- Remove unused code
- Remove nil branch that is never reached
- Correct various ASS tags
- Fix visibility factor using shuffle button

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -1835,7 +1835,6 @@ local function window_controls()
 
     local lo
     local controlbox_w = window_control_box_width
-    local titlebox_w = wc_geo.w - controlbox_w
     local controlbox_left = wc_geo.w - controlbox_w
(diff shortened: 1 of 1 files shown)


Samillion:
I just realized something. If I say, this code is filled with ass, it would be completely appropriate. 😆

nekoxuee:
Since this pr is about cleanup and don't know where to comment this, also noticed that `local titlebox_w` is declared but never used in line 1838

> I just realized something. If I say, this code is filled with ass, it would be completely appropriate. 😆

😭

Samillion:
> Since this pr is about cleanup and don't know where to comment this, also noticed that `local titlebox_w` is declared but never used in line 1838

Damn, that is a nice catch! Thank you.

PR #485, opened 2026-03-04
refactor: improve independant zones show/hide logic

Samillion:
Based on https://github.com/mpv-player/mpv/pull/17496

Kasper re-based our implementation in a more efficient way, which is genuinely nice.

I edited the code for the feature to implement their method instead, modifying it to work with ModernZ.

Obligatory ping to those who tested the feature when we first had it:
/cc @IAmHal9000 @Keith94 @Xurdejl

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -578,7 +578,6 @@ local state = {
     tick_timer = nil,
     tick_last_time = 0,                     -- when the last tick() was run
     hide_timer = nil,
-    wc_hide_timer = nil,
     cache_state = nil,
     idle = false,
     enabled = true,
@@ -752,6 +751,18 @@ end
 
 local function mouse_hit(element)
     return mouse_hit_coords(get_element_hitbox(element))
+end
+
+local function mouse_in_area(names)
+    if type(names) == "string" then names = {names} end
+    for _, name in pairs(names) do
+        for _, cords in pairs(osc_param.areas[name] or {}) do
+            if mouse_hit_coords(cords.x1, cords.y1, cords.x2, cords.y2) then
+                return true
+            end
+        end
+    end
+    return false
 end
 
 local function limit_range(min, max, val)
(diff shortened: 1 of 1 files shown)


nekoxuee:
Tested and works as expected.

As a side note, when `showonpause=false` and `keeponpause=true`, the osc still hides when unhovering while paused. TBH, I find these two options a bit confusing and I'm not sure what the intended use case is for both existing. (it was working like this before, just noticed now while testing)

Samillion:
> As a side note, when `showonpause=false` and `keeponpause=true`, the osc still hides when unhovering while paused. (it was working like this before, just noticed now while testing)

Nice catch. I'll address this in a different PR after this gets merged.

> TBH, I find these two options a bit confusing and I'm not sure what the intended use case is for both existing.

**Intended use**:
- `showonpause=yes` flashes OSC on pause, but then it hides normally.
  - Useful when you want to glance at osc when paused for title, seek location, duration...etc.
- `keeponpause=yes` is basically the same, but as long as it's paused, the osc is there, doesn't hide.

Difference being: hide/show behavior.

Keith94:
I ran into a bug if you define `MOUSE_BTN0 cycle pause` in your input.conf.

1. Open a video
2. Double left click to enter fullscreen
3. Left click on a window controls button

The click goes through the button and pauses the video

Samillion:
Hopefully that fixes it.

I genuinely have no idea how you came up with that test scenario lol, very impressive.

Samillion:
Should I rename `independent_wc` user option to `independent_zones` for clarity?
```lua
independent_zones = false,         -- show/hide top and bottom bar independently based on hover area
```

Or:

```lua
zone_hover = false,               -- show/hide top and bottom bar independently based on hover area
```

Keith94:
independent_zones seems good.

Keith94:
Should it default to true? Same as what mpv will do.

Samillion:
> Should it default to true? Same as what mpv will do.

Definitely. Just clearing things up first.

I'll make it enabled by default in the other PR when I rename the option.

Samillion:
Was the issue fixed by the way? It seems so on my end.

But even before the fix, it wasn't consistent, so I wanted to make sure.

Keith94:
Yeah it looks totally fixed now.

Samillion:
Thank you very much for the thorough testing and confirmation.

I'm definitely going to leave this PR open longer than usual to make sure everything is working as it should.

PR #494, opened 2026-03-07
feat: show bottom bar only (keeponpause + independent_zones)

Samillion:
**Changes**:
- When `keeponpause=yes` and `independent_zones=yes`, only show/keep bottom bar.
- Top bar behaves normally, show/hide based on hover.

image


Keith94:
Much needed change.. Works great.

PR #495, opened 2026-03-07
fix: keep osc visible with button held

Keith94:
Fixed an issue where osc would hide when button/seekbar was held down and `osc_keep_with_cursor` was disabled.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -3718,7 +3718,7 @@ local function render()
         if state[showtime_key] == nil or get_hidetimeout() < 0 then return end
         local timeout = state[showtime_key] + (get_hidetimeout() / 1000) - now
         if timeout <= 0 and get_touchtimeout() <= 0 then
-            if (state.active_element == nil and not mouse_in_area(input_areas)) or not user_opts.osc_keep_with_cursor then
+            if (state.active_element == nil and not mouse_in_area(input_areas)) or (state.active_element == nil and not user_opts.osc_keep_with_cursor) then
                 hide_fn()
             end
         else


Samillion:
You are on fire! Very nice catch. Merging.

On a completely unrelated note: remember when I said I want to reduce redundancy? Well. I did with those PRs that got merged already.... BUT, then I went wild.

The terrifying thing is... it works. Reduced as much as I can. But I have this feeling, I broke everything at the same time, but the universe is not letting me see it lol.

Keith94:
@Samillion There is also another issue when `osc_keep_with_cursor` is disabled...

Essentially, the cursor autohide timer only seems to start "working" _after osc hides_, rather than at the normal time (i.e. respecting `cursor-autohide` value). This means your cursor is stuck on screen longer than expected after osc hides.

I have no idea how to fix that. There is a property ("cursor-autohide") but I don't know if that's helpful.

Keith94:
> You are on fire! Very nice catch. Merging.
> 
> On a completely unrelated note: remember when I said I want to reduce redundancy? Well. I did with those PRs that got merged already.... BUT, then I went wild.
> 
> The terrifying thing is... it works. Reduced as much as I can. But I have this feeling, I broke everything at the same time, but the universe is not letting me see it lol.

Hmm I doubt you would have broken anything (if so, we should be able to catch it later anyways). Always happy to test your changes.

Samillion:
> Essentially, the cursor autohide timer only seems to start "working" _after osc hides_, rather than at the normal time (i.e. respecting `cursor-autohide` value). This means your cursor is stuck on screen longer than expected after osc hides.

Most likely it's due to tick request, which essentially re-inits everything, including mouse position. Never thought it would affect cursor hide. I'll look into it.

Keith94:
Thank you!

I should mention, the issue only happens if your cursor is hovering a button while the osc fades out.
If not touching an button, cursor autohide works as expected.

PR #496, opened 2026-03-07
refactor: add helpers to reduce redundancy and improve logic

Samillion:
**Changes**:
- Add helpers to reduce redundancy, so instead of using the same code over and over, call the helper instead
  - `get_ytdl_format()` 
  - `strip_empty_args(args) `
  - `wc_hoverstyle(color)`
  - `setup_bg_elements(posX, posY, osc_w, osc_alpha3, wc_alpha3)`
  - `build_cache_seek_ranges()`
- Improve logic for `contains(list, item)` helper
- Remove redundant `or` inside `set_osc_styles()`
- Improve logic for `limit_range(min, max, val)`
- Improve logic for `is_image()` and move it to related section
- Various small logic improvements to eliminate branching

@Keith94 I either did it right, or did it so wrong even my fridge will break lol.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -487,16 +487,17 @@ local function set_osc_locale()
 end
 
 local function contains(list, item)
-    local t = type(list) == "table" and list or {}
-    if type(list) ~= "table" then
+    local t
+    if type(list) == "table" then
+        t = list
+    else
+        t = {}
         for str in string.gmatch(list, '([^,]+)') do
             t[#t + 1] = str:match("^%s*(.-)%s*$") -- trim spaces
         end
     end
     for _, v in ipairs(t) do
-        if v == item then
-            return true
-        end
+        if v == item then return true end
     end
     return false
 end
@@ -522,9 +523,9 @@ end
 local osc_styles
 
 local function set_osc_styles()
-    local playpause_size = user_opts.playpause_size or 28
-    local midbuttons_size = user_opts.midbuttons_size or 24
-    local sidebuttons_size = user_opts.sidebuttons_size or 24
+    local playpause_size = user_opts.playpause_size
+    local midbuttons_size = user_opts.midbuttons_size
+    local sidebuttons_size = user_opts.sidebuttons_size
     osc_styles = {
         osc_fade_bg = "{\\blur" .. user_opts.fade_blur_strength .. "\\bord" .. user_opts.fade_alpha .. "\\1c&H0&\\3c&H" .. osc_color_convert(user_opts.osc_color) .. "&}",
         window_fade_bg = "{\\blur" .. user_opts.window_fade_blur_strength .. "\\bord" .. user_opts.window_fade_alpha .. "\\1c&H0&\\3c&H" .. osc_color_convert(user_opts.osc_color) .. "&}",
@@ -767,12 +768,7 @@ local function mouse_in_area(names)
 end
 
 local function limit_range(min, max, val)
-    if val > max then
-        val = max
-    elseif val < min then
-        val = min
-    end
-    return val
+    return math.max(min, math.min(max, val))
 end
 
 -- translate value into element coordinates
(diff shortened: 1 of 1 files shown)


Keith94:
Initially everything checks out.

I'm not even sure if it's related, but the script is sometimes allowing me to fully drag the seekbar with the right mouse button held down, which was causing a lot of erratic behavior with snapping to nearest chapter etc.


https://github.com/user-attachments/assets/37dd43d0-8590-4cb1-b60b-d2034e3336d8

Samillion:
> with the right mouse button held down

Huh? how?

I can't drag at all with right click, only works to seek for nearest chapter on click, not drag/held.

Hmmmm

Samillion:
This is only happening with this PR?

Samillion:
Ok, I think I found how to replicate this.

If I right click and hold then after hold left click and drag, it shows what you showed me.

But that's because I'm basically holding both buttons at the same time, so input is racing between left click behavior or right click behavior.

Is that what you meant?

Keith94:
You may be right. Earlier it seemed to happen with a left click drag -> release -> right click drag. But its's very inconsistent.

Keith94:
Either way, I doubt most users are going to face whatever weirdness I encountered there.. lol
I can test the PR out more thoroughly later.

Samillion:
> Either way, I doubt most users are going to face whatever weirdness I encountered there.. lol 

I can't think of why it would happen with this PR. I'll check pre-PR behavior as well.

If it's not related and just a general behavior we've had, I'll try to think of a way to handle that racing issue. Hopefully lol.

> I can test the PR out more thoroughly later.

Much appreciated ❤️

Keith94:
I can't find anything wrong with this. Fade options etc. are all working. Good job!

Samillion:
Thank you so much for testing.

This PR essentially doesn't change much code. It merely puts repeated codes into helpers then calls them as they were.

So instead of : (pseudo logic)

```
List
- Item 1
- Item 2
- Item 3

List
- Item 1
- item 2
- Item 3
```

It would be:

```
function build_list()
    List
    - Item 1
    - item 2
    - Item 3
end

build_list()
build_list()
```

Then various small logic changes which should result in the same behavior. This is merely to avoid branching.

```
local var

If val == true then
    var = true
else
    var = false
end
```

Would be:
```
local var = (val == true)
```

The one thing I'm unclear about is the seek thing you mentioned. I can't still figure out if it's caused by this PR or if it's a pre-existing race condition.

Samillion:
> I'm not even sure if it's related, but the script is sometimes allowing me to fully drag the seekbar with the right mouse button held down, which was causing a lot of erratic behavior with snapping to nearest chapter etc.
> dVJYCUPo08.mp4

Are you able to get the same behavior without/before this PR?

Keith94:
I'll check again tonight. My guess is preexisting or bug from previous forks, since the PR is mainly a refactor like you said.

Keith94:
So I can definitely reproduce it in master branch. Nothing to do with your PR.
Tomorrow I can try in mpv itself or older forks.. and maybe find out the cause?

Samillion:
You are awesome, thanks again for always testing thoroughly. <3

Samillion:
> Initially everything checks out.
> 
> I'm not even sure if it's related, but the script is sometimes allowing me to fully drag the seekbar with the right mouse button held down, which was causing a lot of erratic behavior with snapping to nearest chapter etc.
> dVJYCUPo08.mp4

I think I found the cause. `state.playing_and_seeking` which is used with `mouse_seek_pause` option.

Right now it's being used inside `ne.eventresponder["reset"]`, which triggers on any `up` event for any button (ie: left or right click)

Still trying to work it out though.

Keith94:
Very interesting.. I'll try to confirm a fix, but can't guarantee. It was very inconsistent getting it to trigger at all.

PR #497, opened 2026-03-07
refactor: remove seek_resets_hidetimeout option

Keith94:
Makes it so `reset_timeout()` depends on `osc_on_seek` instead of `seek_resets_hidetimeout`. It seems more intuitive that timeout always resets if user wants to see osc when seeking.


Samillion:
You are 100% correct. Well done.

Merging.

PR #499, opened 2026-03-07
refactor: remove hide_empty_playlist_button option

Keith94:
Removed the `hide_empty_playlist_button` option. It was redundant given the `playlist_button` option does the same thing.

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -103,8 +103,7 @@ Create `modernz.conf` in your mpv script-opts directory:
 | volume_control             | yes           | show mute button and volume slider                                                                                 |
 | volume_control_type        | linear        | volume scale type: `"linear"` or `"logarithmic"`                                                                   |
 | playlist_button            | yes           | show playlist button: Left-click for simple playlist, Right-click for interactive playlist                         |
-| hide_empty_playlist_button | yes           | hide playlist button when no playlist exists                                                                       |
-| gray_empty_playlist_button | yes           | gray out the playlist button when no playlist exists                                                               |
+| gray_empty_playlist_button | no            | gray out the playlist button when no playlist exists                                                               |
 | download_button            | yes           | show download button on web videos (requires yt-dlp and ffmpeg)                                                    |
 | download_path              | ~~desktop/mpv | default download directory for videos. [Learn more about setting paths here](https://mpv.io/manual/master/#paths). |
 | screenshot_button          | no            | show screenshot button                                                                                             |
--- a/modernz.conf
+++ b/modernz.conf
@@ -142,8 +142,6 @@ volume_control=yes
 volume_control_type=linear
 # show playlist button: Left-click for simple playlist, Right-click for interactive playlist
 playlist_button=yes
-# hide playlist button when no playlist exists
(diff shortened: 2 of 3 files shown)


Samillion:
Question (not a rejection, genuinely asking). Do you think these scenarios might be asked for by users?

```EditorConfig
[default]
    profile-desc=Some default options
    script-opts-append=modernz-playlist_button=yes
    script-opts-append=modernz-hide_empty_playlist_button=no

[Media]
    profile-desc=Apply options for Movies and TV Shows directories
    profile-cond=get("path", ""):lower():match("\\movies\\") or get("path", ""):lower():match("\\tv shows\\")
    profile-restore=copy-equal
    script-opts-append=modernz-playlist_button=yes
    script-opts-append=modernz-hide_empty_playlist_button=yes

[Slideshow]
    profile-desc=Apply options for slideshow directories
    profile-cond=get("path", ""):lower():match("\\slides\\")
    profile-restore=copy-equal
    script-opts-append=modernz-playlist_button=no
```

That's why I said I'm genuinely asking, as I'm not sure if this feature is worth removing or not.

Keith94:
You're right, I think there's an appeal to have it show/hide dynamically. I forgot about that part lol. Thanks for explaining.

Keith94:
I realized that you could get the same behavior as `hide_empty_playlist_button` with this conditional profile.

```editorconfig
[autohide-playlist]
	profile-desc=Show playlist button only when playlist exists
	profile-cond=p["playlist-count"] > 1
	profile-restore=copy-equal
	script-opt=modernz-playlist_button=yes
```

So do you still think a dedicated option is needed for something like that?

Samillion:
I genuinely have no idea. Same issue with the window controls in fullscreen PR. https://github.com/Samillion/ModernZ/pull/557

There are two points fighting each other in this:
- Redundancy in general should be removed. If it can be done already, it shouldn't be added
- Xurdejl conclusion that "if the feature isn't enabled or obvious, it doesn't exist" point is absolutely correct

There are so many options in mpv, that make practically anything possible. There is literally a game made that runs in mpv and mpv as a pdf viewer lol.

At the same time, it means there are so many options that the average users just never know they exist in mpv. 

Saying "can be done with a profile" is easy for for those that know, but to those that don't, it all sounds Martian to them.

I honestly have no idea what is better.

Samillion:
Which also applies to https://github.com/Samillion/ModernZ/pull/550

I am torn., can't decide.

Keith94:
I feel that if something is not super essential (feature/bug fix) then it should not be added unless there is lots of demand.

For example, `visibility_audioonly` option was removed from PR https://github.com/mpv-player/mpv/pull/17504 because it was suggested we can use autoprofile instead.

>Which also applies to #550 

Indeed.. however my PR would make those buttons follow the same logic as audio, subtitles and volume bar, which is: hide the buttons when they're not applicable. I suppose we need to choose one behavior and apply it to all these buttons.

nekoxuee:
I think part of the “too many options” problem comes from turning every change into a user configurable option out of fear of complaints.

There is a reason why so many things are opinionated, they’re usually backed by usability studies and experience.

Samillion:
Fully agreed.

PR #507, opened 2026-03-08
refactor: remove `osd-msg` from volume and speed

Keith94:
Since the speed button and volume slider update in real time, I don't think osd-msg is necessary. It adds extra noise to the screen.


Samillion:
Fully agreed. In fact, the volume bar one shouldn't be there in the first place. Because the way we have scroll wheel behavior on it specifically uses `no-osd`.

Well done. Merging.

Keith94:
I didn't even notice.
Since you brought it up, maybe its better that wheel up/down uses osd-msg after all? Because otherwise it's vague what the volume changes are.

PR #509, opened 2026-03-09
feat: add indication of muted state to volume tooltips

Keith94:
- Change volume button tooltip to "Mute/Unmute" to show what it does (English only)
- Append (Muted) to volume bar tooltip if audio is muted.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -374,7 +374,7 @@ local language = {
         subtitle = "Subtitle",
         no_subs = "No subtitles available",
         no_audio = "No audio tracks available",
-        muted = " (Muted)",
+        muted = "Muted",
         playlist = "Playlist",
         no_playlist = "Playlist is empty",
         chapter = "Chapter",
@@ -2863,7 +2863,7 @@ local function osc_init()
         local volume = mp.get_property_number("volume", 0) or 0
         -- show only one decimal, if decimals exist
         volume = volume % 1 == 0 and string.format("%.0f", volume) or string.format("%.1f", volume)
-        return state.mute and (volume .. locale.muted) or volume
+        return state.mute and (volume .. " (" .. locale.muted .. ")") or volume
     end
     ne.eventresponder["mbtn_left_up"] = command_callback(user_opts.vol_ctrl_mbtn_left_command)
     ne.eventresponder["mbtn_right_up"] = command_callback(user_opts.vol_ctrl_mbtn_right_command)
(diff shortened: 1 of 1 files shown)


Samillion:
If I may suggest:
```lua
    ne.slider.tooltipF = function(pos)
        if audio_track_count <= 0 then return end
        local vol = set_volume(pos)
        return state.mute and (vol .. " (Muted)") or vol
    end
```

Instead of:
```lua
    ne.slider.tooltipF = function(pos)
    if audio_track_count > 0 then
        local vol = set_volume(pos)
        if state.mute then
            vol = vol .. " (Muted)"
        end
            return vol
        end
    end
```

The other part I disagree on, because it removes a useful functionality.
```lua
ne.tooltipF = function() return user_opts.tooltip_hints and (state.mute and locale.unmute or locale.mute) or "" end
```

Right now, the volume button indicates current volume level, which changes on scroll wheel up/down while hovering it.

Very useful in vertical videos where volume bar wouldn't be showing, or if we ever decide to show/hide volume bar as a user option. (ie: only show volume button)

Keith94:
Thanks for that advice. Will fix all that later.

Samillion:
Last note: The hardcoded `Muted` should be a locale so it changes when a different language is used.

Keith94:
Made some changes. Let me know if you like.

Samillion:
If I may suggest again, for consistency:
```diff
-muted = " (Muted)",
+muted = "Muted",
```

```diff
-return state.mute and (volume .. locale.muted) or volume
+return state.mute and (volume .. " (" .. locale.muted .. ")") or volume
```

Then we need to remember to add the muted phrase to locale JSON, which we can do since we have a locale issue and PR coming soon.

Keith94:
Yeah I was trying to write it like that but was having a skill issue.

Samillion:
> but was having a skill issue

You're genuinely doing great. I know that you're starting out, but you actually have a grasp and do changes frequently now.

Well done, don't sell yourself short.

Keith94:
Thanks bro. 😁

Please wait, I'll update a bit later.

Samillion:
Take all the time you need, no rush at all.

Keith94:
Please check the latest. Should be fine now

Samillion:
Terribly sorry about the delay, I had to leave the house again.

Changes look good, well done. When I come back I'll test and merge.

Samillion:
Awesome, thank you very much.

Merging.

PR #510, opened 2026-03-09
refactor: add osd-msg for volume wheel up/down

Keith94:
Makes it more clear on volume changes.


Samillion:
Agreed. Thank you very much.

Merging.

PR #511, opened 2026-03-09
fix: improve playing_and_seeking state use to avoid race on drag

Samillion:
**Fixes**: https://github.com/Samillion/ModernZ/pull/496#issuecomment-4016080054

**Changes**:
- Clear `playing_and_seeking` state in `mbtn_left_down` events on seekbar slider
- Move `mouse_seek_pause` branch to `mbtn_left_up` instead of `reset` event responder
  - `reset` event responder triggers on any `up` event, which could cause an early unpause

This avoids any stuck state which could cause erratic behavior on multiple button presses (up + down, right + left). I hope lol.

@Keith94

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -3202,14 +3202,22 @@ local function osc_init()
     end
     ne.eventresponder["mbtn_left_down"] = function (element)
         element.state.mbtnleft = true
+        state.playing_and_seeking = false  -- clear state
         mp.commandv("seek", get_slider_value(element), "absolute-percent+exact")
     end
     ne.eventresponder["shift+mbtn_left_down"] = function (element)
         element.state.mbtnleft = true
+        state.playing_and_seeking = false
         mp.commandv("seek", get_slider_value(element), "absolute-percent")
     end
     ne.eventresponder["mbtn_left_up"] = function (element)
         element.state.mbtnleft = false
+        if state.playing_and_seeking then
+            if mp.get_property("eof-reached") == "no" and user_opts.mouse_seek_pause then
+                mp.commandv("cycle", "pause")
+            end
+            state.playing_and_seeking = false
+        end
     end
     ne.eventresponder["mbtn_right_down"] = function (element)
         local chapter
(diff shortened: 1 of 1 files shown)


Keith94:
I can still reproduce, but the steps on getting it to happen are not clear. Specifically, dragging seekbar handle with only right click is still possible.

Samillion:
> Specifically, dragging seekbar handle with only right click is still possible.

You mean right click hold + (then/and) left click hold + drag? 
If yes, are there any erratic behaviors or just that you're able to drag?

Or just able to drag by holding right click and nothing else? <= This I can't reproduce.

Keith94:
>You mean right click hold + (then/and) left click hold + drag?
If yes, are there any erratic behaviors or just that you're able to drag?

Yes, I can still do this, but I'm only able to drag. No weird behavior,

>Or just able to drag by holding right click and nothing else?

This happens to me too but it's hard to reproduce. Likely not worth investigating.

Samillion:
Oh, so the erratic behavior is fixed? As in, no race condition, just being able to drag?

I'll try it with a touchpad later, maybe it triggers differently there than a regular mouse. I can't drag at all with a right click only.

Keith94:
Sorry I should clarify. The original video https://github.com/Samillion/ModernZ/pull/496#issuecomment-4016080054 is me dragging with RMB only. That's the only time the "snapping to chapter" behavior in the video happens for me while I'm dragging.

If I hold two buttons at the same time then there doesn't appear to be any race condition, as far as I can tell. Just a normal seek drag occurs.

I'm also using a mouse with it. No touchpad.

Samillion:
I'm terribly sorry, I'm having a hard time understanding, please bare with me.

What is the bug before the PR? Just being able to drag?

With this PR, any bugs or unexpected behavior other than being able to drag?

> I'm also using a mouse with it. No touchpad.

You're able to click and drag with just right click and nothing else before or during?

Keith94:
I'll do some testing and explain more in a little bit.

Keith94:
Hai I found a regression with this... I'll open a new bug?

Keith94:
>What is the bug before the PR? Just being able to drag?

>You're able to click and drag with just right click and nothing else before or during?

I'll try to reproduce and address these in the new bug.

Issue #513, opened 2026-03-09
Respect playback state after dragging seekbar


InterestingFallacy:
### mpv Information

```bash

```

### Reproduction Steps

When in paused state and seekbar is dragged, playback is unpaused after letting go of the seekbar. This is ok if playback was already at unpaused state but not if it was paused. User may want to seek to a different position and not immediately want playback to start. Existence of `mouse_seek_pause` may make this a bit tricky but should be doable maybe? 

### Expected Behavior

Respect pause state after dragging seekbar

### Actual Behavior

Playback always resumes after dragging seekbar

### I carefully read all instructions and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information.

Samillion:
Good point. Will fix soon.

Thank you for reporting.

Samillion:
Fix PR is up, will merge soon.

PR #514, opened 2026-03-09
fix: respect playback state after dragging seekbar

Samillion:
Fixes: #513 

**Changes**:
- Keep track of pause state when a drag event occurs, to respect playback state after event ends with `mouse_seek_pause=yes`

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -3202,18 +3202,21 @@ local function osc_init()
     end
     ne.eventresponder["mbtn_left_down"] = function (element)
         element.state.mbtnleft = true
+        element.state.was_paused = mp.get_property("pause") == "yes"
         state.playing_and_seeking = false  -- clear state
         mp.commandv("seek", get_slider_value(element), "absolute-percent+exact")
     end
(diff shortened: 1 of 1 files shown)


nekoxuee:
I was trying to find what did I do to fix this a few months ago, but can't find the changes because I did a single commit with a lot of changes and I don't remember.

Doing it this way seems to work, and the icon doesn't change while paused and dragging. Don't know if there are any drawbacks.

```diff
- state.playing_and_seeking = true
-    if mp.get_property("pause") == "no" and user_opts.mouse_seek_pause then
+ if not state.playing_and_seeking and mp.get_property("pause") == "no" and user_opts.mouse_seek_pause then
+    state.playing_and_seeking = true
```

InterestingFallacy:
Change in PR works. 
@Xurdejl's change crashes script when applied on top of current Master.

Samillion:
> @Xurdejl's change crashes script when applied on top of current Master.

Out of curiosity, could you share the trace for the crash? I'm not at home right now so I'm unable to test.

If it's not a bother, of course.

Issue #389, opened 2025-05-26
Sometimes subtitles stay with 80% of subtitle position after being raised by UI


4zjata:
### mpv Information

```bash
C:\Users\Szymon\AppData\Local\MPV>mpv --version
Error parsing option sub-ass-use-video-data (option not found)
C:/Users/Szymon/AppData/Local/MPV/portable_config/mpv.conf:36: setting option sub-ass-use-video-data='aspect-ratio' failed.
mpv v0.38.0-730-g7c672e33 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Aug 30 2024 17:05:55
libplacebo version: v7.349.0 (v7.349.0-9-gefb8934-dirty)
FFmpeg version: N-116796-gb730defd5
FFmpeg library versions:
   libavcodec      61.11.100
   libavdevice     61.2.100
   libavfilter     10.2.102
   libavformat     61.5.101
   libavutil       59.35.100
   libswresample   5.2.100
   libswscale      8.2.100
```
I updated and it still happens

### Reproduction Steps

i have no idea how to reproduce this, it dosent happen all the time and i dont know what i do differently when subs stay raised. Stock osc does not change position of subtitles when appearing, so --no-config helped.

### Expected Behavior

Subs will come back to 100% of position after osc disappears. 

### Actual Behavior

Subs sometimes stay at 80% after osc disappears.

### Log File

[output.txt](https://github.com/user-attachments/files/20438840/output.txt) 

### I carefully read all instructions and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information.

liJie-wk:
<img width="952" alt="Image" src="https://github.com/user-attachments/assets/82de220b-d1e7-4531-9536-2e2e48c8aa47" />
I also have the same problem

xiaoleichen:
This happens when you quit mpv while the osc is visible and subtitle is raised. MPV remembers the raised position of the subtitle next time it opens the same file, regardless of whether the osc is visible.

You can add this line to your **`mpv.conf`** file: `watch-later-options-remove=sub-pos`
It will then not remember the position of subtitles on quit.

Samillion:
Fixed in https://github.com/Samillion/ModernZ/pull/592

If `sub_margins` or `osd_margins` options are used and there is conflict in watch later options, you will now be warned with the following:
[...]

PR #523, opened 2026-03-10
feat: add sub_margins, osd_margins and dynamic_margins options

Samillion:
**Changes**:
- Add `sub_margins` option: Raises subtitles above OSC if it's visible, then restores it when OSC is hidden
- Add `osd_margins` option: Adjusts displayed OSD position to not overlap with OSC/WC when it is visible
- Add `dynamic_margins` option: Allows subtitles and OSD to be adjusted when visibility mode is set to auto (default)
- Reset margins and subtitle position on shutdown (mpv exit)
- Deprecate `raise_subtitles`
  - Also warn user if `raise_subtitles` was used that both `raise_subtitles` and `raise_subtitles_amount` are deprecated, to use `sub_margins` and `dynamic_margins` instead.

Based on https://github.com/mpv-player/mpv/pull/17544 

**Key difference**:
- We use `sub-pos` to adjust subtitles, they use `sub-margin-y`
- `sub-margin-y` does not work with ASS subtitles unless `sub-ass-override=strip` is used, which defeats the purpose of ASS subtitles
- Also, `sub-margin-y` might  be affected by `sub-use-margins`

cc @Keith94 as I know you wanted `osd_margins`

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -78,12 +78,23 @@ Create `modernz.conf` in your mpv script-opts directory:
 | window_title_font_size  | 26               | window title font size                                                    |
 | window_controls         | yes              | show window controls (close, minimize, maximize) in borderless/fullscreen |
 
-### Subtitle display settings
+### Subtitle and OSD display settings
+**IMPORTANT**: It is recommended to add the following in your `mpv.conf`, to circumvent subtitle or OSD bad positioning when watch later options are used.
 
-| Option                | Value | Description                                                            |
-| --------------------- | ----- | ---------------------------------------------------------------------- |
-| raise_subtitles       | yes   | raise subtitles above the OSC when shown                               |
-| raise_subtitle_amount | 125   | amount by which subtitles are raised when the OSC is shown (in pixels) |
+```EditorConfig
+watch-later-options-remove=sub-pos
+watch-later-options-remove=osd-margin-y
+```
+
+If `sub_margins` is enabled, the subtitles are raised above the OSC. Calculated by both `osc_height` and `fade_alpha`, specifically whichever of them is highest.
+
+So, to control the raise amount of subtitles, adjust your `osc_height` and `fade_alpha` options in `modernz.conf` to change the OSC boundary.
+
+| Option          | Value | Description                                    |
+| --------------- | ----- | ---------------------------------------------- |
+| sub_margins     | yes   | raise subtitles above the OSC when shown       |
+| osd_margins     | no    | adjust OSD to not overlap with OSC             |
+| dynamic_margins | yes   | update margins dynamically with OSC visibility |
 
 ### Buttons display and functionality
 


Keith94:
Wow that was quick. Will try it out very soon.

For subtitles, how much margin does it know to offset by? Specifically what I'm wondering is would it still overlap with title or chapter title if enabled? I personally don't use either of these, so it doesn't bother me if they overlap though.

Why sub-pos and sub-margin-y both exist but accomplish the same thing? Now I'm just thinking out loud at this point lol.

Samillion:
> I'm wondering is would it still overlap with title or chapter title if enabled?

It should never overlap. This method uses the margins set by OSC height. similar to how select menu (long playlist) does not overlap with OSC.

> Why sub-pos and sub-margin-y both exist but accomplish the same thing?

We were joking about it on irc at one point, that mpv has so many options, a student can write their thesis on them and pass. 😆

Samillion:
Also, on an unrelated note, this PR has the phrase `ASS subtitles` and it's completely normal. Thank you.

Keith94:
>We were joking about it on irc at one point, that mpv has so many options, a student can write their thesis on them and pass. 😆

You're not wrong lol. Scrolling through the documentation and you can tell it's more than bloated.

What is that shutdown event? Does it reset the script-adjusted sub pos on shutdown?

Samillion:
> What is that shutdown event? Does it reset the script-adjusted sub pos on shutdown?

Ensures that they're reset back to default/user config. In essence, ~~to prevent a scenario like this: https://github.com/Samillion/ModernZ/issues/389~~

Samillion:
I'll update docs and `modernz.conf` after testing this PR for a while.

I need a break from PRs lol.

Keith94:
If WC is visible and you scroll up/down on the volume bar (which triggers OSD) it seems to be preventing the WC from timing out. However I don't know if that's expected so I'll point it out anyway.

Subtitle works well too. Maybe it raises a bit high if the title and/or chapter title is disabled. Should those options be taken into account?

>Ensures that they're reset back to default/user config. In essence, to prevent a scenario like this: https://github.com/Samillion/ModernZ/issues/389

I'm not sure if this is working as intended. When doing a quit-water-later it seems they're permanently raised on next startup.

Why not enable osd_margin by default? It's useful.

Samillion:
> If WC is visible and you scroll up/down on the volume bar (which triggers OSD) it seems to be preventing the WC from timing out. However I don't know if that's expected so I'll point it out anyway.

You mean while hovering WC scroll up/down? hovering WC will keep it showing, regardless of mouse actions.

> Subtitle works well too. Maybe it raises a bit high if the title and/or chapter title is disabled. Should those options be taken into account?
> 

> I'm not sure if this is working as intended. When doing a quit-water-later it seems they're permanently raised on next startup.


True, we might need to document, add in `mpv.conf`:
```EditorConfig
watch-later-options-remove=sub-pos,osd-margin-y
```
> Why not enable osd_margin by default? It's useful.

Fear of what you just mentioned, that especially should only be enabled by user when he's aware they should adjust watch later options.

Keith94:
Ah, it's actually seems unrelated. But I was noticing that WC hideout was not working while scrolling the volume bar, like this:

1. Hover the WC
2. Move your mouse to the volume bar
3. Scroll up/down to change the volume (while hovering the volume bar)

Result is the WC stays on screen until you stop changing the volume.

Anyway,

>This method uses the margins set by OSC height.

What happens if we lower the value of `osc_height` user option? Should the subtitle margin adapt to the new height?

Samillion:
Maybe both `sub_margins` and `osd_margins` should be disabled by default? That way those that enable it, would have read the documented watch later options and know how to deal with it.

> What happens if we lower the value of `osc_height` user option? Should the subtitle margin adapt to the new height?

It does, yes. OSC margins use osc_height and fade_alpha value. Whichever is highest, is the OSC boundary.
```lua
osc_param.video_margins.b = math.max(user_opts.osc_height, user_opts.fade_alpha) / osc_param.playresy
```

> Result is the WC stays on screen until you stop changing the volume.

Nice catch.

Keith94:
>watch-later-options-remove=sub-pos

Unless I messed something up, this option doesn't fix the issue. They are still raised on startup.

1. input.conf -> `q quit-watch-later`
2. Hover the OSC at bottom while subtitles raised.
3. Press q
4. Open the same file again.

Samillion:
> Unless I messed something up, this option doesn't fix the issue. They are still raised on startup.

With `watch-later-options-remove=sub-pos`:
```
# E:\test\Bleach- Sennen Kessen-hen - 38.mkv
start=581.247000
```

Without `watch-later-options-remove=sub-pos`:
```
# E:\test\Bleach- Sennen Kessen-hen - 38.mkv
start=555.138000
sub-pos=81.000000
```

My `mpv.conf` watch later:
```EditorConfig
watch-later-dir="~~/cache/watch_later"
write-filename-in-watch-later-config=yes
watch-later-options-remove=sub-pos
```

Keith94:
>It does, yes. OSC margins use osc_height and fade_alpha value. Whichever is highest, is the OSC boundary.

Then it's working as intended. This would be good to document about the fade_alpha. I personally found the lower limit for osc_height is around 115 otherwise it messed up clicking on OSC elements.

>watch-later-options-remove=sub-pos

Ok, sorry, it works with just this line. However  if you add `,osd-margin-y` to the line then it doesn't work (it doesn't reset the sub-pos properly?)

>True, we might need to document, add in mpv.conf:
watch-later-options-remove=sub-pos,osd-margin-y

Samillion:
> Ok, sorry, it works with just this line. However if you add `osd-margin-y` to the line then it doesn't work (it doesn't reset the sub-pos properly?)

True, turns out `-remove` accepts one option at a time. So it would be:
```EditorConfig
watch-later-options-remove=sub-pos
watch-later-options-remove=osd-margin-y
```

But, from my testing, `osd-margin-y` is never saved. so probably only `sub-pos` one is needed for it.

Samillion:
By the way, the whole `watch-later-options-remove=sub-pos` isn't just this feature PR, before with `raise_subtitles` it had the same behavior. Since they both adjust `sub-pos`

But at least now we should document and tell users what to do.

> Then it's working as intended. This would be good to document about the fade_alpha

I will document this as well.

Keith94:
Can we force enable that option? Or just rely on the user to set up their mpv.conf properly.

Either way, the PR works really well. I'm impressed. Thanks for getting it in so quick.

Samillion:
> Can we force enable that option? Or just rely on the user to set up their mpv.conf properly.

I thought about it, but I don't know if we should. We'd be altering the user's configuration without their knowledge. Even if we tie it to an option that could be disabled.

I'm not saying no, I just don't know if yes is ok lol.

Keith94:
That's fair. 

Side note, is user_opts.md even needed? Since modernz.conf has the same content and updating so many files gets a bit too much.

Samillion:
> so many files gets a bit too much.

It is a pain in the ass and probably is read by 1% of users. 

However, it makes readability easier and we're able to add more detailed notes like the ones I just added about OSC boundary, watch later and so on.

Samillion:
All done. I think this PR is ready to merge, let me know if I missed something please. If it's not a bother.

> Result is the WC stays on screen until you stop changing the volume.

I'll try to fix this in another PR.

Keith94:
Looks good to me too.

Keith94:
>Maybe both sub_margins and osd_margins should be disabled by default? That way those that enable it, would have read the documented watch later options and know how to deal with it.

Forgot to reply sorry. I agree they should be disabled by default to avoid issues **or** just disable `dynamic_margins` and keep the other two enabled by default. Then users can turn it on after reading the warning.

Samillion:
I was discussing it on irc, and guido suggested a great solution:
Read watch later options, if they include OSD margin/sub position, disable ModernZ sub/osd options, then warn they need to do X to enable them.

What do you think?

At least then we can enable it all by default.

Edit:
We don't even need to disable them if they're included in watch later options, we can just warn users about it.

Keith94:
Yeah that seems like a good solution!

Keith94:
You had said that osd-margin-y is never saved in watch later options, so you won't have to really add a check for that, right? Just subtitles.

Samillion:
By default, it's never saved in watch later options. However a user might add it for some reason or in the future mpv could add it to be saved.

It's more of a safeguard than necessity.

Keith94:
>We don't even need to disable them if they're included in watch later options, we can just warn users about it.

Don't you think the warning could get missed for a regular user?. I mean warnings don't show up in the interface except for terminal, right? Either way, whatever you think is best.

Samillion:
Not sure what else I can do. I've included a note of it in `modernz.conf`, in user options documentation and will add it for terminal.

Should I make it display OSD text (`show-text`) for 8 seconds with a warning as well?

Keith94:
Right, that might be good to add. It was reported a few times anyway. Good idea.

Samillion:
Pseudo logic:
```
If watch later options has: osd-margin-y or sub-pos then
    warn user in terminal
    warn user in show-text
    set sub_margins=no
    set osd_margins=no
end
```

Samillion:
This returns watch later options:

```lua
mp.msg.info(mp.get_property("options/watch-later-options"))
```

```
start,speed,pitch,edition,volume,mute,audio-delay,gamma,
brightness,contrast,saturation,hue,deinterlace,vf,af,panscan,
aid,vid,sid,sub-delay,sub-speed,sub-pos,sub-visibility,sub-scale,sub-use-margins,
sub-ass-force-margins,sub-ass-use-video-data,sub-ass-override,secondary-sid,
secondary-sub-delay,secondary-sub-pos,secondary-sub-ass-override,
secondary-sub-visibility,ab-loop-a,ab-loop-b,video-aspect-override,
video-aspect-method,video-unscaled,video-pan-x,video-pan-y,video-rotate,
video-crop,video-zoom,video-scale-x,video-scale-y,video-align-x,video-align-y
```

So we can do something like:

```lua
-- wrap with comma "," to address possible first and last position
local watch_later = "," .. (mp.get_property("options/watch-later-options") or "") .. ","

if watch_later:find(",sub-pos,", 1, true) then
    msg.warn("found sub-pos. y u do dis? remove right meow!")
end

if watch_later:find(",osd-margin-y,", 1, true) then
    msg.warn("found osd-margin-y. y u do dis? remove right meow!")
end
```

Keith94:
>    msg.warn("found osd-margin-y. y u do dis? remove right meow!")

Lol. It must definitely include this 😂

PR #524, opened 2026-03-11
fix: don't let input events reset window controls bar hidetimeout

Samillion:
Fixes: https://github.com/Samillion/ModernZ/pull/523#issuecomment-4035396194 reported by @Keith94, ModernZ's local detective. 

**Changes**:
- Top bar (WC bar) `hidetimeout` should reset only if mouse is hovering it.
  - It shouldn't be affected by input events that reset `hidetimeout` on the OSC with `reset_timeout()`.
- Area elements should only affect the specific area. (ie: seek drag should not reset timeout for top bar)

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -3649,7 +3649,9 @@ local function render()
         if state.pause_osc_locked and showtime_key == "showtime" then return end
         local timeout = state[showtime_key] + (hide_timeout / 1000) - now
         if timeout <= 0 and get_touchtimeout() <= 0 then
-            if state.active_element == nil and (not mouse_in_area(input_areas) or not user_opts.osc_keep_with_cursor) then
+            -- area elements should affect the specific area only. (ie: seekbar drag shouldn't affect top bar)
+            local element_blocks_hide = state.active_element ~= nil and mouse_in_area(input_areas)
+            if not element_blocks_hide and (not mouse_in_area(input_areas) or not user_opts.osc_keep_with_cursor) then
                 hide_fn()
             end
         else


Keith94:
>Fixes: #523 (comment) reported by @Keith94, ModernZ's local detective.

If that's true then you must be the dang genius for fixing everything lol.

I'll check it out tomorrow but I assume it works perfectly.

Keith94:
It fixes https://github.com/Samillion/ModernZ/pull/523#issuecomment-4035396194 but it still doesn't work when dragging seekbar handle.

Samillion:
> It fixes [#523 (comment)](https://github.com/Samillion/ModernZ/pull/523#issuecomment-4035396194) but it still doesn't work when dragging seekbar handle.

I think that's caused by

```lua
mp.observe_property("seeking", "native", function(_, seeking)
    if user_opts.osc_on_seek then
        reset_timeout()
    end
end)
```

I'll test soon, will be back home in an hour or so.

Samillion:
If you can test, try replacing it with:

```lua
mp.observe_property("seeking", "native", function(_, seeking)
    if user_opts.osc_on_seek then
        state.showtime = mp.get_time()
    end
end)
```

Keith94:
Pretty sure it was not affected by osc_on_seek when I last tested it, but you'll have to confirm.

Samillion:
Done. https://github.com/Samillion/ModernZ/commit/1f9a52c99e6af6346174d5d10db6bbf3952baffa

Keith94:
Thanks, will check

Samillion:
Take your time, there is no rush at all.

Keith94:
Okay, I could not reproduce those bugs. However I may have randomly found another one (sorry). If you keep spamming spacebar to cycle pause, the WC timeout is interrupted.

Samillion:
> Okay, I could not reproduce those bugs. However I may have randomly found another one (sorry). If you keep spamming spacebar to cycle pause, the WC timeout is interrupted.

That's expected not a bug. `input.conf` does not affect timeout, unless you have `keeponpause` which will keep osc visible on pause/play spam.

Only OSC inputs (ie: buttons) reset its timeout.

Edit:
`keeponpause` or `showonpause`... any options that trigger show osc with pause/seek state, basically.

Keith94:
Ah, okay. I see what you mean. In that case, you fixed it all. 😄

Samillion:
I tend to reply quickly and misread. I understood you correctly, right?

My brain is at half mass, so I need to re-check lol.

Keith94:
I think so...

I personally did not expect that cycle pause would ever keep WC visible. But like you mentioned if I set these options then it correctly follows the hidetimeout.

```lua
showonpause=no
keeponpause=no
```

PR #532, opened 2026-03-12
refactor: improve margins with volume bar

### LINKED ISSUE #525: Make the volume elements closer together

I think it would be nice to adjust margins so volume slider is more connected to volume toggle.

Something like move volumebar 8 pixels to the left. Example:

Line 2094 start_x = start_x + 20
Line 2123 (auto_hide_volbar and 68 or 0)

Note:
I'll revert part of https://github.com/Samillion/ModernZ/commit/fd4caabd9be917a81933f77793d8939a929f71d7 otherwise tooltip will overlap.

Keith94:
- Adjusts the volume button and slider to appear closer and more like a single element.
- Makes volume width consistent between layouts.
- Removes muted phrase from tooltip to avoid overlapping.

Fixes #525

mpv_b9NGGFC8Qe

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -2091,7 +2091,7 @@ layouts["modern"] = function ()
         lo.geometry = {x = start_x, y = refY - 35, an = 5, w = 24, h = 24}
         lo.style = osc_styles.control_3
         lo.visible = (osc_param.playresx >= 600 - outeroffset)
-        start_x = start_x + 28
+        start_x = start_x + 20
 
         -- Volumebar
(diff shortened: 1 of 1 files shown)


Samillion:
Oooo, so pretty. Looks good to me, tested and looks great.

I actually like that the tooltip was removed. It makes more sense having it on the volume button only.

Let me know when it's ready for merge please. No rush at all.

Keith94:
Perfect. Thanks for testing so fast. I have nothing much else to add.

Issue #536, opened 2026-03-12
Make bottomhover always on and remove the toggle option


Keith94:
### Expected behavior of the wanted feature

https://github.com/Samillion/ModernZ/issues/533#issuecomment-4050791803

Samillion:
Leaving this here as a reminder.

----

So remove `bottomhover` (no rename), then let it all be controlled by `bottomhover_zone` and `tophover_zone`

So essentially, change this

```lua
if user_opts.bottomhover then
```

To this:

```lua
if state.zones_hover then
    -- do hover zone/independent zone stuff
else
    -- show/hide on any mouse movement
end
```

----

- Add `zones_hover` as a state:

```lua
zones_hover = user_opts.bottomhover_zone > 0 and user_opts.tophover_zone > 0,
```

- Replace all `user_opts.bottomhover` with `state.zones_hover`
- Adjust descriptions of affected options

PR #539, opened 2026-03-13
feat: add zones_hover_mode option

Samillion:
Fixes: https://github.com/Samillion/ModernZ/issues/536

**Changes**:
- Add `zones_hover_mode` option to replace both `bottomhover` and `independent_zones` options
  - `bottomhover` and `independent_zones` options are completely removed

**Zone hover modes**:
- `always`: show both top and bottom bars on any mouse movement
- `zones`: show both top and bottom bars when mouse hovers either zone's area
- `independent`: show only the specific bar (top or bottom) when area zone is hovered

cc @Xurdejl @Keith94

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -30,24 +30,23 @@ Create `modernz.conf` in your mpv script-opts directory:
 
 ### OSC behaviour and scaling
 
-| Option                  | Value | Description                                                                      |
-| ----------------------- | ----- | -------------------------------------------------------------------------------- |
-| hidetimeout             | 2000  | time (in ms) before OSC hides if no mouse movement                               |
-| fadeduration            | 200   | fade-out duration (in ms), set to `"0"` for no fade                              |
-| fadein                  | no    | whether to enable fade-in effect                                                 |
-| minmousemove            | 0     | minimum mouse movement (in pixels) required to show OSC                          |
-| bottomhover             | yes   | show OSC only when hovering at the bottom                                        |
-| bottomhover_zone        | 130   | height of hover zone for bottomhover (in pixels)                                 |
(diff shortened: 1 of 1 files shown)


nekoxuee:
I was about to reply to the other comment so I will write it here:

I like the idea of making the values one word. 
Should the option be named `osc_trigger` so it describes what it controls? I initially threw `zones_hover_mode` as a quick example without much thought.

Samillion:
> Should the option be named `osc_trigger` so it describes what it controls? I initially threw `zones_hover_mode` as a quick example without much thought.

I'm fine with any name. It's so difficult naming things lol. Maybe we should ask AI to come up with a meaningful and not confusing name.

> I like the idea of making the values one word.

Yeah, makes it easier for the user.

Keith94:
This is a good solution; works well for me!

Samillion:
In theory, `zones_hover_mode` (with the modes) is self explanatory. Plus, it covers any zones we might add in the future.

For example, if we ever want to do something like uosc's volume bar thing.

As I said though, I'm fine with any name.

Keith94:
Make the mode independent by default like before?

Samillion:
> Make the mode independent by default like before?

After the ModernZ defaults discussion, I decided to check and came to the following conclusions:
- Most users have `title-bar=yes` (unless mpv decides to change its default in the future)
- Most embedded players (ie: Youtube), on hover = show (ie: `always`)

Having it default to `always` will be consistent with common behavior. If user prefers a specific behavior, the option is there to adjust. They decide, basically.

What do you think?

Keith94:
Ya, sounds great to me.

nekoxuee:
Looks good

PR #540, opened 2026-03-13
refactor: remove jump_more from chapter skip buttons

Keith94:
<s>* Right-clicking jump buttons are tied to mouse commands instead of user option.</s>
* Removed jump functionality from chapter buttons since it felt out of place.
* Various docs fixes.

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -104,6 +104,7 @@ So, to control the raise amount of subtitles, adjust your `osc_height` and `fade
 | audio_tracks_button        | yes           | show the audio tracks menu button                                                                                  |
 | jump_buttons               | yes           | show "jump forward/backward 10 seconds" buttons                                                                    |
 | jump_amount                | 10            | change the jump amount in seconds                                                                                  |
+| jump_more_amount           | 60            | change the jump amount in seconds on right click                                                                   |
 | jump_icon_number           | yes           | show different icon for `5`, `10`, or `30` second jumps                                                            |
 | jump_mode                  | relative      | seek mode for jump buttons: `"relative"` or `"exact"`                                                              |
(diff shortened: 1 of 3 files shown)


Samillion:
> * Removed jump functionality from chapter buttons since it felt out of place.
> 
> * Various docs fixes.

Nice catch, well done.

> Right-clicking jump buttons are tied to mouse commands instead of user option.

Isn't this already achievable by `jump_more_amount` option?

I only ask because:

```EditorConfig
jump_more_amount=60 
```

is simpler (in use/adjusting) for the average user than

```EditorConfig
jump_backward_mbtn_right_command=seek -60 relative
jump_forward_mbtn_right_command=seek 60 relative
```

Which remains consistent with the options, as in:

```EditorConfig
jump_amount=10
jump_more_amount=60
jump_mode=relative
```

Keith94:
Simpler I would agree, but this gives user a bit more control. This way you could change the amount and seek type for each direction.

- Right click could use exact seeks while left click uses relative.
- Could be used with seek 0 percent to restart playback or something.

Samillion:
Sorry, I don't mean to disagree, I'm just thinking of what could be requested in the future.

> * Right click could use exact seeks while left click uses relative.
> * Could be used with seek 0 percent to restart playback or something.

Wouldn't those be more fit for `input.conf` though?

I'm only asking to see if it will be reverted immediately due to complaints/confusions and to not lengthen the options list if we can help it.

As in, replacing one option with two options to fit a niche scenario is always a concern.

However, fixing the docs and removing it from chapter buttons I fully agree on.

Keith94:
I do see your point. No worries I'll change it back.

Samillion:
Very nice, never even noticed the old details in docs. Well done.

Merging.

Issue #541, opened 2026-03-13
Tooltip runs out of bounds


Keith94:
### mpv Information

```bash

```

### Reproduction Steps

1. show_title=no
2. show_chapter_title=no
3. Play https://www.youtube.com/watch?v=b1Fo_M_tj6w
4. Hover the beginning of seekbar

### Expected Behavior

Chapter title kept in bounds

### Actual Behavior

The beginning of the chapter title is cut off

### I carefully read all instructions and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information.

nekoxuee:
Currently, to avoid at the edges, the script changes the anchor point based on position. For some reason this is not done in the chapter title.

This could be fixed by doing the same there, but I don't think the anchor shift is a good solution to begin with, so it would be like a patch.

Edit: I could change it to calculate the text width to clamp the position at the edge of the video. But, since the script is being cleaned up, not sure if we want to patch it for now and revisit it once the cleanup is done.

Samillion:
> Currently, to avoid at the edges, the script changes the anchor point based on position. For some reason this is not done in the chapter title.

Indeed. This whole block can be simplified. I am not sure why I re-did math and positioning for chapter title in that position. Will do a patch now for a test.

Samillion:
Test script: [modernz.lua](https://github.com/Samillion/ModernZ/blob/chapter_title_bounds/modernz.lua)
Commit: https://github.com/Samillion/ModernZ/commit/c695d840cb5671060467141adc035e5d9045f0bc

**Changes**:
- Fix chapter title tooltip without thumbfast, `show_title=no` and `show_chapter_title=no`
  - Instead of redoing everything, just use the already existing positioning logic

nekoxuee:
> Indeed. This whole block can be simplified. I am not sure why I re-did math and positioning for chapter title in that position. Will do a patch now for a test.

By the way, I have a fix but taking a different approach on tooltip positioning, is that something worth pursuing or would you rather leave it for later?

Samillion:
> is that something worth pursuing or would you rather leave it for later?

Improvements are always welcome. Do you want me to delete this patch?

nekoxuee:
script: [modernz.lua](https://github.com/Xurdejl/ModernZ/blob/664c53a7c8bf17cb5ce82e58b088bd5a42df1834/modernz.lua)
[...]

PR #543, opened 2026-03-14
refactor: tooltip positioning logic

nekoxuee:
Improved tooltip positioning. Adds estimate_text_width() function using mpv's osd overlay to calculate actual text widths, enabling proper horizontal boundary clamping so tooltips stay within screen edges.

fixes: #541

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -1340,25 +1340,6 @@ local function draw_seekbar_progress(element, elem_ass)
 end
 
 local function render_elements(master_ass, osc_vis, wc_vis)
-    -- when the slider is dragged or hovered and we have a target chapter name
-    -- then we use it instead of the normal title. we calculate it before the
-    -- render iterations because the title may be rendered before the slider.
-    state.forced_title = nil
-
-    -- disable displaying chapter name in title when thumbfast is available
-    -- because thumbfast will render it above the thumbnail instead
-    if thumbfast.disabled then
-        if user_opts.chapter_fmt ~= "no" and state.touchingprogressbar then
-            local dur = mp.get_property_number("duration", 0)
-            if dur > 0 then
-                local ch = get_chapter(state.sliderpos * dur / 100)
-                if ch and ch.title and ch.title ~= "" then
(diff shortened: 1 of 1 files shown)


nekoxuee:
So I’ve been trying combinations to see how it worked before.

It seems the current hovered chapter replaces either the title or the active chapter text, unless both are disabled, in which case it finally falls back to the tooltip.

This replace the title approach was a result mpv stock osc design limitations (thick progress bar where the tooltip is displayed inside), and because early versions of ModernX didn't display chapter names either, this code was adopted. 

However, when po5 added thumbnail support to ModernX, they implemented chapter display as a proper tooltip. Even in the thumbnail version of the stock osc, the title replacement behaviour was kept rather than using the tooltip, likely due to the stock layout.

Now that we have a proper tooltip system, the title replacement feels redundant and inconsistent from a UX perspective. So, do we want to keep this legacy behaviour, or move to a tooltip only approach?

Keith94:
>Now that we have a proper tooltip system, the title replacement feels redundant and inconsistent from a UX perspective. So, do we want to keep this legacy behaviour, or move to a tooltip only approach?

Tooltip only seems way better. It's much easier to follow the tooltip to see content changes compared to the OSC title where it's just a static element.

Samillion:
Agreed. Though both should remain, since we have tooltips disable/enable option. This way, no one will be upset.

Let me know when it's ready to merge.

Keith94:
>Agreed. Though both should remain, since we have tooltips disable/enable option. This way, no one will be upset.

What option do you mean? `tooltip_hints`? That doesn't affect the seekbar tooltip because it's a critical element.

Samillion:
True! Completely forgot about that.

Yup, leaving this up to you guys. My brain is extra pretending to exist today lol. Let me know when it's ready to merge, please.

nekoxuee:
Since the seek bar tooltip is always active, and the old title replacement logic was suppressing it based on unrelated user options. I'm removing the replacement code to keep it tooltip only. 

Ready to merge.

Samillion:

Merging.

PR #503, opened 2026-03-08
refactor: replace speed button icon with text label

nekoxuee:
Also moved some code into a function to make it easier to read. Initially I thought about rounding the number but seems like mpv already handles that.

PR #545, opened 2026-03-14
refactor: show .0 for whole number playback speeds

Keith94:
For example, 1× => 1.0× to keep it consistent with fractional speeds.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -3088,8 +3088,8 @@ local function osc_init()
     --tog_speed
     ne = new_element("tog_speed", "button")
     ne.content = function()
-    local speed = mp.get_property_number("speed", 1)
-        return (speed % 1 == 0) and string.format("%.1f×", speed) or string.format("%g×", speed)
+        local speed = mp.get_property_number("speed", 1)
+        return string.format(speed % 1 == 0 and "%.1f×" or "%g×", speed)
     end
     ne.visible = (osc_param.playresx >= visible_min_width)
     ne.tooltip_style = osc_styles.tooltip


Samillion:
This was intentional, no? https://github.com/Samillion/ModernZ/pull/503

cc @Xurdejl 

Edit:
I merely ask because I don't use loop options anywhere, so I don't know what the expected behavior should be.

Samillion:
It does look consistent with other playback speed UIs. I'll let you guys decide that, since I'm not a loop user.

By the way, you can condense your format logic like this if you want:

```lua
    ne.content = function()
        local speed = mp.get_property_number("speed", 1)
        return string.format(speed % 1 == 0 and "%.1f×" or "%g×", speed)
    end
```

Keith94:
I will admit I like to use speed_button_scroll=0.1, so the reasoning behind this PR is the button looks much nicer because it stays in one position without jumping around on certain speed changes.

At the script default value of 0.25 it will still have a similar issue as before though. Unless it can be fixed altogether by aligning text to the left somehow..

Samillion:
It looks really nice, I have no objection whatsoever. I've only cc'd Xurdejl to get his input, since he seems to have went above and beyond into UI comparison and different media players behavior.

My condensing logic didn't remove anything you did, just altered so instead of two `string.format`, to just one and fixed `local speed` indentation.

nekoxuee:
It was intentional, but comes down to personal preference. I don't mind either way, so feel free to change it.

Samillion:
So, merge? Objectively speaking this or current is better?

Like I said, I ask because I have no expectation of behavior since I don't use loop options anywhere.

PR #547, opened 2026-03-15
fix: don't show osc when file looping occurs

### LINKED ISSUE #546: OSC appears on file loop if osc_on_seek=yes
### mpv Information

```bash

```

### Reproduction Steps

1. `osc_on_seek=yes`
2. `loop_button=yes`
3. Play a file with loop enabled
4. Let playback finish so file loops around

### Expected Behavior

OSC does not show on file loops

### Actual Behavior

OSC shows up when file restarts on loop.

Don't know if it's even possible to detect a situation like this.

### Log File

_No response_

Keith94:
Fixes #546

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -3828,12 +3828,11 @@ mp.observe_property("chapter-list", "native", function(_, list)
     request_init()
 end)
 mp.register_event("seek", function()
-local pos = mp.get_property_number("time-pos", 0)
     if state.new_file_flag then
         state.new_file_flag = false
         return
     end
-    if pos ~= 0 and user_opts.osc_on_seek then
+    if user_opts.osc_on_seek and not (state.file_loop and mp.get_property_number("time-pos", -1) == 0) then
         show_osc()
     end
 end)


Samillion:
Hmm. This will not show osc, if user seeks back to 0 (start), while having `osc_on_seek` enabled, right?

Wouldn't adding this instead solve the loop issue?

```lua
mp.observe_property("eof-reached", "bool", function(_, val)
    if val and state.file_loop then
        state.new_file_flag = true
    end
end)
```

Keith94:
>This will not show osc, if user seeks back to 0 (start), while having osc_on_seek enabled, right?

Ah, you are correct..

>Wouldn't adding this instead solve the loop issue?

No, I don't think it's fixed, at least not for me.

Samillion:
I'll be back home in about 2 hours. Terribly sorry about the delay, I'll try to brainstorm with you to come to a solution.

By the way, the seek to 0 thing, I don't mind it. I'm just not sure what the users who have that option will think.

Samillion:
Hmmmm, also you bring a good point in the issue post:

> In log file, shows that looping always causes seek.
> 
> ```
> [  17.631][v][lavf] queuing seek to 0.000000
> [  17.631][v][lavf] execute seek (to 0.000000 flags 32)
> [  17.631][v][lavf] seek done
> [  17.631][v][cplayer] hr-seek, skipping to 0.000000
> ```

Looping back is a seek event to `0`. It would be consistent with the description of `osc_on_seek` I mean. It should show, right?

Genuinely asking, not implying. I don't use this feature so I don't have a perspective of any expected of behavior.

Keith94:
No, I personally wouldn't expect it to show because I did not initiate a seek myself. For example, just look at YouTube on desktop behavior where the file loops cleanly without any popups.

Samillion:
> No, I don't think it's fixed, at least not for me.

By any chance you have `osc_on_start` enabled?

Keith94:
> > No, I don't think it's fixed, at least not for me.
> 
> By any chance you have `osc_on_start` enabled?

Nop, it's disabled.

Samillion:
I think your implementation was the correct one after all, just needs a bit of adjusting. I think. hope. My brain pretends to exist.

```lua
mp.register_event("seek", function()
    if state.new_file_flag then
        state.new_file_flag = false
        return
    end
    if user_opts.osc_on_seek and not (state.file_loop and mp.get_property_number("time-pos", -1) == 0) then
        show_osc()
    end
end)
```

Keith94:
I had faith, but it still won't work with seek to beginning. xD

`home seek 0 absolute-percent`

Samillion:
So the decision to make here:
- Accept that loop back is a seek event, so show osc with `osc_on_seek`
- Seek `0` never shows osc, regardless of `osc_on_seek` status

Otherwise, we'd have to think of a way to differentiate `seek 0` event if it was handled by loop or user, and that is no trivial task.

Keith94:
Obviously if you use a command like this it will work...

`home seek 0.000001 absolute-percent`

Which is a good workaround we can recommend if a user complains about it. Realistically, this is still way better than the current behavior.

Samillion:
Sounds good to me. Do you mind adjusting it to the example I've given?

Explanation:
- `-1` default value ensures that if the value cannot be read, it'll return false. (`-1 == 0`)
- Behavior of `osc_on_seek` will only be altered if file loop is enabled

Keith94:
Done, thank you

PR #548, opened 2026-03-15
fix: don't alter window decoration on pinned state changes

### LINKED ISSUE #544: Pin button and window border

I run with `border=yes` so having the border removed on left clicking pin button seems unexpected.
Right click keeps the border but the left click is a more natural interaction (same as all other buttons).

A solution to this would be to either:

1. Make the left and right clicks customizable/swappable
2. Leave border untouched and let user modify the border state with their own conditional profile.

Keith94:
* Remove border changes on changing pinned state.
* Also added an OSD message when it's toggled to stay true to other button behavior.

Fix #544

If you want to remove the border when using the pin button, use this [auto-profile](https://mpv.io/manual/master/#conditional-auto-profiles). [(thanks to Sam)](https://github.com/Samillion/ModernZ/issues/544#issuecomment-4062125549)

```editorconfig
[Pin-ontop]
    profile-desc=Options when ontop (pin) is enabled
    profile-cond=ontop == true and not fullscreen
    profile-restore=copy-equal
    border=no
```

### DIFF
--- a/extras/pip-lite/README.md
+++ b/extras/pip-lite/README.md
@@ -18,7 +18,7 @@ To adjust them you can either:
 | `geometry_restore`  | `50%:50%`     | window position when PiP is disabled. default: center [[details](https://mpv.io/manual/master/#options-geometry)]      |
 | `geometry_delay`    | `0.02`        | window position delay (in seconds) to avoid any racing issues                                                          |
 
-You can easily control other ModernZ or mpv options when PiP is enabled by using an auto-profile, for example:
+You can easily control other ModernZ or mpv options when PiP is enabled by using an auto-profile in `mpv.conf`, for example:
 
 ```EditorConfig
 [Window-PiP]


Samillion:
Beautiful, well done. Should we add `border=no` to the profile in `pip-lite`?
https://github.com/Samillion/ModernZ/tree/main/extras/pip-lite

and maybe in `pip-lite` as well:

```diff
-You can easily control other ModernZ or mpv options when PiP is enabled by using an auto-profile, for example:
+You can easily control other ModernZ or mpv options when PiP is enabled by using an auto-profile in `mpv.conf`, for example:
```

Keith94:
Should be updated. Please check.

Samillion:
Lmao you fixed opacity typo too, thank you very much. I shall hide in a corner with a thousand shames from my ancestors.

Keith94:
Oh my, that is overkill lol. No problem

PR #550, opened 2026-03-15
feat: auto hide track and chapter buttons when none available

Keith94:
For context, this PR make these buttons follow the same logic as audio, subtitles and volume bar, which is: hide the buttons when they're not applicable.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -2021,7 +2021,7 @@ layouts["modern"] = function ()
     local audio_track = state.audio_track_count > 0
     local subtitle_track = state.sub_track_count > 0
     local jump_buttons = user_opts.jump_buttons
-    local chapter_skip_buttons = user_opts.chapter_skip_buttons and mp.get_property_number("chapters", 0) > 0
+    local chapter_skip_buttons = user_opts.chapter_skip_buttons and type(state.chapter_list) == "table" and next(state.chapter_list)
     local track_nextprev_buttons = user_opts.track_nextprev_buttons
     local fullscreen_button = user_opts.fullscreen_button
     local info_button = user_opts.info_button


Samillion:
Terribly sorry about the delay. I'm preparing for the release to not overwhelm one release with many changes.

After the cached observe is merged, I'll release, then we can work on everything else.

I have a long to do list, mostly about reducing code repetition.

Keith94:
No problem, take your time. I can see you've been busy tidying up and optimizing etc. Awesome job btw.

Later we can always discuss the PR and what direction to go in.

Samillion:
Now that things have calmed down, I can finally come back to check lol.

Thank you very much for your effort, it works great.

Question: Should these be tied to an option or do you think the behavior doesn't need one? 

I think your implementation is good as is, to be honest, adding options would complicate things, right?

Keith94:
Oh hai. I think it's good without having an option. It keeps it consistent with other buttons which nobody complained about.

If a user cannot survive with the new behavior they always have an autoprofile option.

Samillion:
Indeed. I swear this "what if user" stuff is going to give me an ulcer.

Keith94:
Yeah, it's hard to know what users might expect. Even though it's based on osc.lua doesn't mean it has to follow it word-for-word.

Samillion:
Indeed, that's why I rely on feedback. I'm stuck in a position in the middle between code alteration and average user. I've seen both sides go way overboard when it comes to altering the simplest of things.

User(s) freak out on the simplest of change, programmers freak out on the simplest of change. In the end it's taxing because almost always it turns into a defense against person, not issue at hand.

As I'm sure you've seen in many issues and PRs (mpv or otherwise), the struggle to define personal preference vs what's actually good or would benefit the experience is a very thin line.

Samillion:
Would this need grouping or does lua handle this correctly? It's probably fine.

```diff
-ne.visible = pl_count > 1 or contains(user_opts.buttons_always_active, "playlist_prev") and (osc_param.playresx >= pl_nav_visible_w)
+ne.visible = (pl_count > 1 or contains(user_opts.buttons_always_active, "playlist_prev")) and (osc_param.playresx >= pl_nav_visible_w)
```

Keith94:
Mpv project is the same way that they rely on feedback. They are not like a web browser that has metrics built in to see exact percentage of how much a feature is used/clicked on etc.

>Would this need grouping or does lua handle this correctly? It's probably fine.

Not sure, I saw no issues. But please adjust if needed.

Samillion:
Thank you for adjusting, at least we can feel at ease that the logic will be consistent and not freakout due to a weird lua parse.

Question: Instead of calling the chapters property, should we use the chapter state?

```lua
local chapter_skip_buttons = user_opts.chapter_skip_buttons and mp.get_property_number("chapters", 0) > 0
```

Though I'm not sure what's the best way, to measure length or use [`next()`](https://www.lua.org/manual/2.4/node31.html):

Length:

```lua
local chapter_skip_buttons = user_opts.chapter_skip_buttons and #state.chapter_list > 0
```

next()

```lua
local chapter_skip_buttons = user_opts.chapter_skip_buttons and next(state.chapter_list) ~= nil
```

Samillion:
Length would be fine, since we're safe guarding it anyway by making sure it's never nil. right? Unless I'm missing something.

Keith94:
I can change it to that if you prefer.

Samillion:
Sorry, I didn't mean to make it sound like a demand/request. I was genuinely asking (performance wise) which would be better.

Keith94:
The `chapters` property seems like it's more lightweight call, no? Either way it's probably similar.

Samillion:
Thank you very much.

Merging.

PR #555, opened 2026-03-17
feat: add speed button font size option

Keith94:
Speed button size can be customized, making it more compatible with various fonts. Currently, the font size can look too small depending on the applied font.

This also makes it follow side button colors for consistency.


Samillion:

Merging.

PR #557, opened 2026-03-18
feat: add windowcontrols_fullscreen option

nekoxuee:
From: https://github.com/mpv-player/mpv/commit/5999676

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -28,7 +28,6 @@ local user_opts = {
 
     idlescreen = true,                     -- show mpv logo when idle
     window_top_bar = "auto",               -- show OSC window top bar: "auto", "yes", or "no" (borderless/fullscreen)
-    windowcontrols_fullscreen = false,     -- show window controls in fullscreen
     showwindowed = true,                   -- show OSC when windowed
     showfullscreen = true,                 -- show OSC when fullscreen
     showonselect = false,                  -- show OSC when a select menu is open
@@ -81,6 +80,7 @@ local user_opts = {
     window_title = "${media-title}",       -- same as title but for window_top_bar
     window_title_font_size = 26,           -- window title font size
     window_controls = true,                -- show window controls (close, minimize, maximize) in borderless/fullscreen
+    windowcontrols_fullscreen = true,      -- show window controls in fullscreen
 
     -- Subtitle and OSD display settings
     sub_margins = true,                    -- raise subtitles above the OSC when shown


Samillion:
Beautiful, thank you very much. Could you enable it by default, please? Once they disabled that, they have been bombarded by complaints on irc.

The only reason that PR was made and disabled by default is because <ins>**one person**</ins> (not PR author) in mpv wanted it.

**Side note**:
Without this PR, it is doable by a conditional profile. 

```EditorConfig
[Fullscreen-WC]
    profile-desc=Disable window controls in fullscreen
    profile-cond=fullscreen
    profile-restore=copy-equal
    script-opts-append=modernz-window_controls=no
```

Or simply `window_controls=no` in `modernz.conf` if they have `title-bar=yes` in `mpv.conf`

Not a rejection, just merely asking if do you think we'd need it. I'm fine either way.

nekoxuee:
I didn't know about the irc stuff, I downloaded a client the other day but barely join. 

Sure I can change the default value but if you want to keep it as a conditional profile I don't have any strong opinion in this feature so up to you.

Samillion:
I don't mind it either, I just find it redundant. However, that's my personal bias, so I never reject an idea just because I don't use it.

I genuinely never know what the general perspective is from the user's point of view. I can always say "yeah, you can do that already", but then a response might be: "well, an option is easier" lol.

Samillion:
Adding the option will be consistent with "based on osc.lua". Just enabling it by default would suffice.

Edit:
It should be listed in `### Title bar settings` below `window_controls` I think.

nekoxuee:
Sorry it was quite late and couldn't get back to this until now

After reading the original pr discussion and checking how other programs behave (at least on Windows), it makes sense to have the option, especially since mpv kept it disabled by default.

Samillion:
No worries at all, I always forget I can edit the files myself, sorry about that.

Again, thank you very much for the effort.

Samillion:
https://github.com/mpv-player/mpv/pull/17602

nekoxuee:
I understand both points of view. Still think hiding it is the "correct" behaviour, but I get people have simply grown used to the old behaviour after all these years.

The same thing happened when I removed the leading zeros from the time codes. They are objectively useless and just slow down readability with visual noise, but someone preferred the clutter and an option was added.

That said, I don’t mind the current default given that people are used to it, and I respect the decision.

Samillion:
Indeed. It's definitely one of those "old habits die hard" kind of thing. It can be argued which is better for a month with no conclusion because in reality it's personal preference not actual general benefit/loss.

The only key argument that wins here from mpv irc is this:

> It was like this for years and we never got reports of having window controls removed or disabled. However, once we disabled them, users started to show up on irc to complain.

This is the only factor or difference between the two behaviors.

Like you though, I don't mind it either way and will respect the final decision, I can workaround with or without the option.

Issue #297, opened 2025-01-07
Vertical video mode: activation areas of play-pause button and time code overlap too much.


xiaoleichen:
### mpv Information

```bash
mpv 45c1c58 Copyright © 2000-2024 mpv/MPlayer/mplayer2 projects
 built on Nov 18 2024 18:32:19
libplacebo version: v7.349.0
FFmpeg version: 7.1
FFmpeg library versions:
   libavcodec      61.19.100
   libavdevice     61.3.100
   libavfilter     10.4.100
   libavformat     61.7.100
   libavutil       59.39.100
   libswresample   5.3.100
   libswscale      8.3.100
```

### Reproduction Steps

* Open a vertical video and make sure time codes are above play-pause button.
* Click lower half of the time codes

### Expected Behavior

Only time codes should respond to the click.

### Actual Behavior

Both time codes and play-button are highlighted, but randomly sometimes play-pause button responds, and other times time codes responds. 

Play pause button responds:
https://github.com/user-attachments/assets/a893cab8-4ca3-4196-8c13-5f80a1385cab

Time code responds:
https://github.com/user-attachments/assets/e532c2cc-356d-4a58-b40a-9885e67c525e

### I carefully read all instructions and confirm that I did the following:

- [X] I tested with the latest mpv version to validate that the issue is not already fixed.
- [X] I provided all required information.

Samillion:
Follow up here, please: https://github.com/Samillion/ModernZ/issues/297

Samillion:
Follow up here, please: https://github.com/Samillion/ModernZ/issues/295

Samillion:
Should be fixed now.

Issue #295, opened 2025-01-05
Is it possible to adjust the position of title or seekbar tooltip?


jiangzhenjerry:
### Expected behavior of the wanted feature

For context, I'm using a 27" 4K monitor and I set the followings:
```
vidscale = "",
scalewindowed = 1.5,
scalefullscreen = 1.5,
```

For this setup, the minimum font size (for any element) for my comfortable viewing is 18. Therefore I change the `tooltip` font size to 18 (it used to follow `time_font_size` but it's separate now, which is fine and not the focus of this request.) However, the top of the seekbar tooltip overlaps with the title (Screenshot 1). Even if I change the tooltip font size to 16, it still looks off (Screenshot 2). And apparently, I'm too old for the default value of 14.

Can you point me as to where I can either move the tooltip down a little or move the title up a little? That'd make my day. Thanks.

Screenshot 1, where tooltip font size is set to be 18:
<img width="362" alt="1" src="https://github.com/user-attachments/assets/99cb7299-0d50-4027-a90b-9c57b03fd0d8" />

Screenshot 2, where tooltip font size is set to be 16:
<img width="362" alt="2" src="https://github.com/user-attachments/assets/76f3d1b1-2741-4c82-b99a-1611c6897270" />

Samillion:
Adjust title height here:
https://github.com/Samillion/ModernZ/blob/743f45d63863b9a9508136e7da67f0bbb42e6bd6/modernz.lua#L1722-L1723

- The `102` height refers to when a chapter title exists
- `90` with no chapters (like your screenshots)

I've been trying to think of a way to scale for font size, like adding additional height/width. I can't think of something solid. Hopefully I will soon.

I'll try to look up used methods instead of re-inventing the wheel.

Samillion:
There is a solution where it allows users to control, inspired by your request here, but I'm not sure if it's a viable one to be honest.

Adding height and width offset user options for elements that are affected by font size, so that you can adjust then in `modernz.conf` and it won't be affected by updates (ie: you don't have to change them every time you update).

**Examples:**
```EditorConfig
title_height=y
title_with_chapter_height=y
chapter_title_height=y
# ...etc
```

Samillion:
CC @Keith94 as he might be interested as well

Samillion:
> the `tooltip` font size to 18 (it used to follow `time_font_size`

Self reminder: add this back if the solution above is applied, or consider adding a `tooltip_font_size` option

[...]

PR #558, opened 2026-03-18
refactor: remove code repetition in buttons and tooltip style

Samillion:
This is the last change to prepare for v0.3.1 release. A considerable size refactor to remove code repetition for tooltips and OSC/WC buttons.

**Changes**:
- Unify time code width use in `modern` and `modern-compact` with `get_time_codes_width()`
- Remove repeated tooltip style assignment, instead use it once in `new_element()`
- Add `wc_button()` to avoid code repetition for window control buttons
- Add `left_side_button()` and `right_side_button()` to avoid code repetition with OSC buttons in `modern` layout
- Add `compact_right_side_button()` to avoid code repetition with OSC buttons in `modern-compact` layout
- Remove unused `tooltip_left_offset` option

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -230,7 +230,6 @@ Useful when adjusting font size or type, this will help you change the affected 
 | time_codes_height           | 35    | time codes height position                                             |
 | time_codes_centered_height  | 57    | time codes height position with portrait window                        |
 | tooltip_height_offset       | 2     | tooltip height position offset                                         |
-| tooltip_left_offset         | 5     | if tooltip contains many characters, it is moved to the left by offset |
 | portrait_window_trigger     | 1000  | portrait window width trigger to move some elements                    |
 | hide_volume_bar_trigger     | 1150  | hide volume bar trigger window width                                   |
 | notitle_osc_h_offset        | 25    | osc height offset if title above seekbar is disabled                   |


Samillion:
@Keith94 @Xurdejl I apologize for the ping. I'd really appreciate any help in testing with this one, if it's not a bother. Any input on the changes is more than welcome.

Keith94:
I tested various scenarios and works perfectly, and the code looks really nice and simplified.

I don't want to give you a rushed answer, so I could do more testing in the evening.

Keith94:
Ah, I don't know if it's related to the font I was using, but I came across this issue: https://github.com/Samillion/ModernZ/issues/297

Can you reproduce that?

Samillion:
I can't apologize enough for asking you guys to test, it's a burden that should be on me. However, I seem to miss a few test scenarios that I never think of, and this is a significant refactor.

It's not even done yet lol, but I'm leaving the rest for the release after. (center buttons, modern-compact left buttons....etc).

> Ah, I don't know if it's related to the font I was using, but I came across this issue: #297
> 
> Can you reproduce that?

I tested now, can't reproduce. If I had to guess, it would be font size. If you adjusted size, you can increase the height offset for timecodes in options.

One of the main reasons I switched back to Roboto font 😭

Keith94:
I see. So with the new get_time_codes_width() it doesn't adapt to non-default font sizes? If you use high values for example.

Samillion:
> I see. So with the new get_time_codes_width() it doesn't adapt to non-default font sizes? If you use high values for example.

Only determines an estimated width, unfortunately. Font scale (height), has to be a manual change with the options. It's why we added these options: (from https://github.com/Samillion/ModernZ/issues/295)

```EditorConfig
# Elements Position
# Useful when adjusting font size or type

# title height position above seekbar
title_height=96
# title height position if a chapter title is below it
title_with_chapter_height=108
# chapter title height position above seekbar
chapter_title_height=91
# time codes height position
time_codes_height=35
# time codes height position with portrait window
time_codes_centered_height=57
# tooltip height position offset
tooltip_height_offset=2
# if tooltip contains many characters, it is moved to the left by offset
tooltip_left_offset=5
# portrait window width trigger to move some elements
portrait_window_trigger=1000
# hide volume bar trigger window width
hide_volume_bar_trigger=1150
# osc height offset if title above seekbar is disabled
notitle_osc_h_offset=25
# osc height offset if chapter title is disabled or doesn't exist
nochapter_osc_h_offset=10
# seek hover timecodes tooltip height position offset
seek_hover_tooltip_h_offset=0
# osc height without offsets
osc_height=132
```

Keith94:
`tooltip_left_offset=5` is this still used in the code?

Samillion:
> `tooltip_left_offset=5` is this still used in the code?

Nice catch, nope, doesn't exist and should be removed. Thank you.

nekoxuee:
Should the time codes could be moved to the same position of the compact layout (far right above the progressbar) instead of being on top of the play button when there is no space?

Samillion:
> far right above the progressbar instead of being on top of the play button when there is no space?

Good idea. We can address that after v0.3.1 release for sure.

nekoxuee:
> Ah, I don't know if it's related to the font I was using, but I came across this issue: https://github.com/Samillion/ModernZ/issues/297

I mentioned that because of this. Otherwise everything else looks good to me.

Keith94:
There is a minor hit box issue with the time codes if it's worth fixing. Besides that it seems fine.

https://github.com/user-attachments/assets/431949a6-7da4-4724-aa34-cfa0d437c832

Samillion:
> There is a minor hit box issue with the time codes if it's worth fixing. Besides that it seems fine.

The time width function definitely needs improvements. I'll see what can be done about it.

Thank you very much for testing.

Issue #567, opened 2026-03-21
Thicker border on tooltip


Keith94:
### Expected behavior of the wanted feature

I had an idea of increasing the border so the tooltip is more readable, especially against other white text.

Something like this. What do you think?

```lua
tooltip = "{\\blur0\\bord2.5\\1c&HFFFFFF&\\3c&H000000&\\fs" .. user_opts.tooltip_font_size .. "\\fn" .. user_opts.font .. "}",
```

## Proposed

<img width="530" height="199" alt="Image" src="https://github.com/user-attachments/assets/87da0dc3-b845-4cee-b8d1-f1d3243d6113" />

## Default

<img width="539" height="203" alt="Image" src="https://github.com/user-attachments/assets/ade232ff-f5ed-408c-876b-169dc461065b" />

krisu5:
Purposed opacity on tooltip's text border makes it harder to read than with 100% opacity IMO (still better then current one though). Still I personally agree on this suggestion.

Maybe tooltip border should be customizable in these ways:
- Border size
- Color
- Opacity.

Keith94:
>Opacity.

That seems to be controlled by this line. Removing it gets rid of it entirely.

https://github.com/Samillion/ModernZ/blob/8eaf2905d96e6a6a8ec5890442b6cf7161f6e469/modernz.lua#L1500

>Border size
Color

We can potentially make these follow the already existing thumbnail options:

    thumbnail_border_color = "#FB8C00",    -- color of the border for thumbnails (with thumbfast)
    thumbnail_border = 3

<img width="367" height="202" alt="Image" src="https://github.com/user-attachments/assets/fa4f65cf-ade6-4c14-b645-2ea70ead89cc" />

Would that be a good solution?

krisu5:
> Would that be a good solution?

Yeah sure, as long as you don't need thumbfast 😄

Samillion:
This is why we had current hovered chapter replace chapter title (if it was enabled) or title (if chapter title was disabled)

But the behavior was removed here: https://github.com/Samillion/ModernZ/pull/543

Should I add it back?

PS. @krisu5 Big fan of Aegis icons. ❤️

Keith94:
Mm, adding a border should be a better solution because it improves the contrast in all situations. A lot of video players have a similar thing with their tooltips.

It can also be customized to user's preference.

Keith94:
@Samillion Are you against the change or should I make a PR for it?

[...]

PR #569, opened 2026-03-21
feat: wrap seek bar tooltips with a box, similar to thumbnail

Samillion:
Fixes: https://github.com/Samillion/ModernZ/issues/567

**Changes**:
- Wrap time and chapter seek bar tooltips with a box, same as thumbnail.
- Only draw tooltip box for seekbar
- Scale with tooltip font size, not time stamp font size
- Increase spacing between seek bar tooltips
- Adjust tooltip opacity, corner radius and horizontal padding
- Change default value of `seek_hover_tooltip_h_offset` to `5`
- Add `tooltip_seek` to `osc_styles`

image

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -1498,7 +1498,7 @@ local function render_elements(master_ass, osc_vis, wc_vis)
                             if chapter_text and osd_w and r_w > 0 then
                                 local titleY = ty - tooltip_fs - 2 * pad_v - 5
                                 elem_ass:new_event()
-                                elem_ass:append("{\\rDefault\\alpha&H80}")
+                                elem_ass:append("{\\rDefault\\alpha&H80&}")
                                 elem_ass:pos(tx - chapter_width / 2 - pad_h, titleY - tooltip_fs - pad_v)
                                 elem_ass:an(7)
                                 elem_ass:append(osc_styles.tooltip_seek)
@@ -1553,7 +1553,7 @@ local function render_elements(master_ass, osc_vis, wc_vis)
                                 if chapter_text then
                                     local chapterY = thumbY * r_h - thumbPad * r_h - pad_v - 5
                                     elem_ass:new_event()
-                                    elem_ass:append("{\\rDefault\\alpha&H80}")
+                                    elem_ass:append("{\\rDefault\\alpha&H80&}")
                                     elem_ass:pos(tx - chapter_width / 2 - pad_h, chapterY - tooltip_fs - pad_v)
                                     elem_ass:an(7)
                                     elem_ass:append(osc_styles.tooltip_seek)
@@ -1573,7 +1573,7 @@ local function render_elements(master_ass, osc_vis, wc_vis)
                         -- tooltip label background box
                         if element.name == "seekbar" then
                             elem_ass:new_event()
-                            elem_ass:append("{\\rDefault\\alpha&H80}")
+                            elem_ass:append("{\\rDefault\\alpha&H80&}")
                             elem_ass:pos(tx - tooltip_width / 2 - pad_h, ty - tooltip_fs - pad_v)
                             elem_ass:an(7)
                             elem_ass:append(osc_styles.tooltip_seek)


Keith94:
Hmm, if you increase the UI scale above 1, then the tooltip background starts to increasingly run away off screen edge.
This is what 3 looks like.

image

Samillion:
Is that an issue though? Genuinely asking. The text positioning is correct, right? At the far edge of the slider (last pos)

Keith94:
> Is that an issue though? Genuinely asking. The text positioning is correct, right? At the far edge of the slider (last pos)

The text is correct.

Using `local pad_h, pad_v = 0, 3` fixes that padding issue. Might be biased, but I would make it default to avoid potential issues lol.

Samillion:
> Using `local pad_h, pad_v = 0, 3` fixes that padding issue. Might be biased, but I would make it default to avoid potential issues lol.

`0` only when no thumbfast or for all scenarios?

Keith94:
With thumbfast you would run into the same issue. Unless you would rather add a user option.

Samillion:
Would `pad_h` value of `2` work? I don't have videos with long chapters on hand at the moment.

What I mean is, does `pad_h` have to be `0`? On reasonable scales.

Keith94:
>Would pad_h value of 2 work?

Yeah, pad_h of 2 looks perfect up to scale values of around 1.4 on my video. At 1.5 scale it starts to clip the edge.
But that should be a good solution.

Samillion:
Done. Lowered horizontal padding to 2 from 5.

Keith94:
Thank you. How can we run it borderless? Border outline is always enabled, right?

Samillion:
Done.

image

Keith94:
That's awesome. Such an improvement over how it looked yesterday haha.

Btw, if you like the border, I'm not against having it available. I just wanted a way to disable lol. What you have now is good for sure tho.

Samillion:
> Btw, if you like the border, I'm not against having it available. I just wanted a way to disable lol. What you have now is good for sure tho.

I really like your idea of no border, makes it blend in, looks right.

I'll merge tomorrow ish to see if @Xurdejl has suggestions, since he suggested the tooltips to be wrapped like that.

I'll have to re-do a lot of preview images and video 😭 but I won't do it right away, because I saw a sneaky Xurdejl messing with purple variants for seekbar lol.

Edit:
Speaking of purple. I genuinely love the orange default colors. I have no idea why, but it looks nice and unique ish.

Keith94:
If you don't mind more suggestions, should tooltip match the osc color of #000000 ? If not at least be a bit darker so it's easier to read.

Right now it follows `thumbnail_border_color` but it may not be super obvious that it's tied to tooltip. Should it be renamed?

>I genuinely love the orange default colors. I have no idea why, but it looks nice and unique ish.

Of course, orange is always sexy, but I'm very biased in that regard lol.

Samillion:
> If you don't mind more suggestions, should tooltip match the osc color of #000000 ? If not at least be a bit darker so it's easier to read.

Good idea. Done.

image

Keith94:
The radius is at 8 but what do you think about 12. More pill shaped like that.

Samillion:
Sorry, before I address that, do we still need `seek_hover_tooltip_h_offset` option?

That was initially added to adjust height in case of overlap. With this, overlap is fine, right? In other words, should the option be removed?

Keith94:
The only way it's gonna overlap now is if you push `hover_button_size` to some extreme value.

But hover size  ≤ 115 is perfect for most cases anyway. Should be good to remove.

Samillion:
> The radius is at 8 but what do you think about 12. More pill shaped like that.

`12` feels a bit too much, no? even the round corner ends with a pointy center.

nekoxuee:
> I'll merge tomorrow ish to see if @Xurdejl has suggestions, since he suggested the tooltips to be wrapped like that.

Looks good. 

> I'll have to re-do a lot of preview images and video 😭 but I won't do it right away, because I saw a sneaky Xurdejl messing with purple variants for seekbar lol.

I wasn't really. I just don't remove old stuff from the conf file

> Edit: Speaking of purple. I genuinely love the orange default colors. I have no idea why, but it looks nice and unique ish.

No need to change them, orange is ok

>> The radius is at 8 but what do you think about 12. More pill shaped like that.
>
> 12 feels a bit too much, no? even the round corner ends with a pointy center.

if you want to make it pill shaped what about `(fs + 2 * pad_v) / 2)`

Samillion:
> if you want to make it pill shaped what about `(fs + 2 * pad_v) / 2)`

Ah, to scale it with tooltip font size to ~10 radius? I'm ok with that if you guys think it looks better.

Keith94:
> > if you want to make it pill shaped what about `(fs + 2 * pad_v) / 2)`
> 
> Ah, to scale it with tooltip font size to ~10 radius? I'm ok with that if you guys think it looks better.

Can you make a before and after demo. Ideally it would look closer to `slider_rounded_corners` appearance.

Samillion:
> Can you make a before and after demo. Ideally it would look closer to `slider_rounded_corners` appearance.

I'm out right now, but can post previews when I'm back home in a couple of hours.

Without changing tooltip font size, setting radius to a static `10` would showcase the result of `((fs + 2 * pad_v) / 2)`

Keith94:
> > Can you make a before and after demo. Ideally it would look closer to `slider_rounded_corners` appearance.
> 
> I'm out right now, but can post previews when I'm back home in a couple of hours.
> 
> Without changing tooltip font size, setting radius to a static `10` would showcase the result of `((fs + 2 * pad_v) / 2)`

Okay, that looks like a good compromise.

Samillion:
@Keith94 @Xurdejl so, yes to `((fs + 2 * pad_v) / 2)`?

nekoxuee:
I said that instead of static 10 because it creates a perfect pill regardless of font size, in the case of wanting a pill shape of course

Keith94:
> @Keith94 @Xurdejl so, yes to `((fs + 2 * pad_v) / 2)`?

Looks good to me!

Samillion:
Done. Let me know if anything else needs adjusting. 

I won't rush merging this so that we can have our time gawking at pills. 😆

Issue #392, opened 2025-06-06
Finish Time Indicator


arfadex:
### Expected behavior of the wanted feature

Something similar to this:

![Image](https://github.com/user-attachments/assets/eefeaf27-1e28-41a4-8b08-82c175491458)

PR #574, opened 2026-03-23
feat: add configurable clock display

mouadlotfi:
## Summary
- add a configurable clock display beside the playback speed control in the modern layout and centered in modern-compact
- show the estimated end time by default, with click-to-toggle current time display
- expose `show_clock` and `clock_font_size` in `modernz.conf`

### DIFF
--- a/modernz.conf
+++ b/modernz.conf
@@ -95,6 +95,10 @@ time_font_size=16
 tooltip_font_size=14
 # speed button font size
 speed_font_size=16
+# show "Ends at" time next to speed button
+show_clock=yes
+# font size of the "Ends at" display
+clock_font_size=16
 
 # Title bar settings
 # show window title in borderless/fullscreen mode
--- a/modernz.lua
+++ b/modernz.lua
@@ -119,6 +119,9 @@ local user_opts = {
     speed_button_click = 1,                -- speed change amount per click
     speed_button_scroll = 0.25,            -- speed change amount on scroll
 
+    show_clock = true,                     -- show "Ends at" time next to speed indicator
+    clock_font_size = 16,                  -- font size of the "Ends at" display
+
     loop_in_pause = true,                  -- enable loop with mouse actions on pause button
 
     buttons_always_active = "none",        -- force buttons to always be active. can add: playlist_prev, playlist_next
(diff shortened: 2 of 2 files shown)


Keith94:
The idea has been rejected before: https://github.com/Samillion/ModernZ/issues/392.

This clock also breaks the cache info display. Maybe it's better to show it as an optional tooltip instead.

Samillion:
Hmmm. Is this a desirable feature? I honestly haven't seen it anywhere. Genuinely asking, not rejecting.

For the most part I always found showing remaining time to be an equivalent of this, however, I'd love to know if there is a use case that would benefit from showing the exact time to end instead of a count down. I'm very curious.

mouadlotfi:
> Hmmm. Is this a desirable feature? I honestly haven't seen it anywhere. Genuinely asking, not rejecting.
> 
> For the most part I always found showing remaining time to be an equivalent of this, however, I'd love to know if there is a use case that would benefit from showing the exact time to end instead of a count down. I'm very curious.

I've seen this in players like Jellyfin and I find it useful. It complements a countdown rather than replaces it.

Samillion:
I don't mind it, but the PR needs to be rebased because of conflicts.

**Issues**:
- Why is it tied to the speed button? You want it to show only if the speed button is showing as well?
- Why was the speed button placement moved?
- `show_clock` should be disabled by default
- For clock text width, use a wide template, ie: `88:88`. Having a bit more is better than less.
- Consider @Keith94 's suggestion to show as a tooltip instead? To avoid layout breakage as described

Keith94:
From what I saw online, seems like Jellyfin puts the clock element above the seekbar rather than below it. Dunno what implications that might have on osc height and offsets, etc.

Having a tooltip for timecodes sounds reasonable as long as the contents update in real time, but the toggle behavior would need adjusting.

Samillion:
> Having a tooltip for timecodes sounds reasonable as long as the contents update in real time, but the toggle behavior would need adjusting.

It should be a tooltip for time stamps not seekbar, right? `0:00 / 0:00`

Keith94:
> > Having a tooltip for timecodes sounds reasonable as long as the contents update in real time, but the toggle behavior would need adjusting.
> 
> It should be a tooltip for time stamps not seekbar, right? `0:00 / 0:00`

Yup, that's what I meant to say. If tooltip was indeed considered as a solution, it could also depend on user_opts.timecurrent state to show different tooltips perhaps (such as current time display / estimated end time mentioned in OP).

Keith94:
So it could look like this as a tooltip to avoid cluttering up the interface.

image


mpv_cJ1kl2lXGC

mouadlotfi:
What about this?

image

Samillion:
Sure, but you'll have to consider the following:
- In `modern` layout, if a video is portrait, the time stamps move to where you want the clock to show
- In `modern-compact` layout, time stamps exist there permanently

Samillion:
Closing due to no response.

Whenever you're ready, please don't hesitate to either re-open the PR and rebase or make a new PR.

If you need any help implementing this, please let us know.

PR #576, opened 2026-03-23
fix: exclude speed button from hover size effect

Keith94:


Samillion:
Terribly sorry about the delay.

Question (not rejection):
Should this be excluded? We initially excluded the others because they're long strings of text which made it look odd on hover.

Since speed button is in theory still a button, that's why I ask.

Keith94:
Valid point. Technically it's a situational thing depending on how long your decimal values get.

The cache also has a speed element which does not change size so this felt like an oversight somehow.

Oh well. The decision is fully in your hands. Either way is fine.

Samillion:
> The cache also has a speed element which does not change size so this felt like an oversight somehow.

Indeed. I think the reason we excluded that because as a whole it's a long string.

I don't mind it either way, I'm trying to think which would be aesthetically best for the speed button. Hence why I said it's not a rejection, more of a brainstorming.

Keith94:
Does anybody technically use long speed like 1.23456? If so I could instantly see the appeal of disabling. I doubt it happens but it's possible.

PR #579, opened 2026-03-24
fix: make slider hover effect work independently

### LINKED ISSUE #578: Allow hover effect for slider only

User option `hover_effect_for_sliders` only works if hover size is enabled for all buttons. I would argue there should be a way for just the slider to have a hover effect without depending on all buttons.

```lua
        -- Apply size hover_effect only if hovering over the handle
        if handle_hovered and user_opts.hover_effect_for_sliders then
            if hover_effect_size then
                rh = rh * (user_opts.hover_button_size / 100)
            end
        end
```

For example,

```diff
---  if handle_hovered and user_opts.hover_effect_for_sliders then
---  if hover_effect_size then
+++ if handle_hovered then
+++ if hover_effect_size or user_opts.hover_effect_for_sliders then
```

Keith94:
Previously, the slider hover effect would not work unless hover size was enabled globally.

- Slider hover toggle now works by itself.
- Add `slider_hover_size` defaulting to 130 for greater effect.
- Rename `hover_button_size` to `button_hover_size` for consistency.

Closes #578


Samillion:
Thank you very much.

PR #581, opened 2026-03-25
feat: make the volume unmute on volume changes

Keith94:
When changing volume with OSC, unmute the audio if it was previously muted. It mimics the behavior of YouTube and Windows volume sliders.

EDIT: If you want it to also affect the mouse wheel, add this to `modernz.conf`:

```editorconfig
vol_ctrl_wheel_down_command=no-osd set mute no; osd-msg add volume -5
vol_ctrl_wheel_up_command=no-osd set mute no; osd-msg add volume 5
volumebar_wheel_down_command=no-osd set mute no; osd-msg add volume -5
volumebar_wheel_up_command=no-osd set mute no; osd-msg add volume 5
```

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -114,7 +114,7 @@ So, to control the raise amount of subtitles, adjust your `osc_height` and `osc_
 | track_nextprev_buttons     | yes           | show next/previous playlist track buttons                                                                          |
 | volume_control             | yes           | show mute button and volume slider                                                                                 |
 | volume_control_type        | linear        | volume scale type: `"linear"` or `"logarithmic"`                                                                   |
-| volumebar_unmute_on_click  | yes           | unmute audio when adjusting volume slider with left click                                                          |
+| volumebar_unmute_on_click  | no            | unmute audio when adjusting volume slider with left click                                                          |
 | playlist_button            | yes           | show playlist button: Left-click for simple playlist, Right-click for interactive playlist                         |
 | hide_empty_playlist_button | no            | hide playlist button when no playlist exists                                                                       |
 | gray_empty_playlist_button | no            | gray out the playlist button when no playlist exists                                                               |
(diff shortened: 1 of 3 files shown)


Keith94:
Oops, I did not change the compact layout. Have to fix that.

Samillion:
Having the unmute in options is a good idea, however hard-coding the function will definitely receive complaints at some points.

If I may ask, why not remove the mute property set and leave it with options only?

Keith94:
> Having the unmute in options is a good idea, however hard-coding the function will definitely receive complaints at some points.
> 
> If I may ask, why not remove the mute property set and leave it with options only?

Oh, the reason is because mouse down on the volume slider is not configurable at all it seems.

I had to add that to mouse down otherwise it would not behave consistently with the other options.

Is there another solution?

Samillion:
I've been hesitating on this one to be honest. The main reason is that we have had comments about changing original state (seek + pause). This would definitely apply here as well.

I might be overthinking it, but it's why I've been hesitant about this. Thoughts?

Keith94:
I added a toggle for it in case of willingness to go back to original state.

Should I disable it by default? I could also revert the key binds and just add the option (default disabled).

Samillion:
Agreed. That way both are satisfied, default behavior is unchanged and those that want to mimic Youtube behavior are able to do so, good idea.

Keith94:
Should be done. Thanks for always being so cool about letting me add stuff.

Samillion:
> Should be done. Thanks for always being so cool about letting me add stuff.

No, thank you for all your effort. The other day @Xurdejl and I were praising you on irc. We even brought up that you could find things we could never think of, a true detective.

Keith94:
Damn, thanks for all the credit guys.
Sometimes I feel annoying but I can't help but pay a LOT of attention to details! 🤣

PR #588, opened 2026-03-27
fix: enable hover effects for gray playlist button

### LINKED ISSUE #586: If playlist button is grayed, hover effects not working
### mpv Information

```bash

```

### Reproduction Steps

1. Put `gray_empty_playlist_button=yes` in config
2. Play a file with no playlist
3. Hover playlist button

### Expected Behavior

Button has hover effects

### Actual Behavior

No hover effects.

Seems to be caused by presence of `element.enabled` in this line
```lua
local hovered = mouse_hit(element) and is_clickable and element.enabled and state.mouse_down_counter == 0
```

### Log File

_No response_

Keith94:
Fixes the issue where the playlist button would have no hover effects if `gray_empty_playlist_button=yes` and playlist is empty.

Closes #586

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -2791,7 +2791,7 @@ local function osc_init()
 
     --playlist
     ne = new_element("playlist", "button")
-    ne.enabled = have_pl or not user_opts.gray_empty_playlist_button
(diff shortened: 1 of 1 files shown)


Samillion:
Thank you very much.

Merging.

PR #591, opened 2026-03-28
refactor: remove tooltips_for_disabled_elements

### LINKED ISSUE #587: Clarify the `tooltips_for_disabled_elements` option

The `tooltips_for_disabled_elements` option was added in PR #134 mainly for the audio and subtitle buttons.

PR #244 removed the buttons when none are available. So the option only works for playlist button (grayed out).
Is keeping this option still necessary when we already have `tooltip_hints`?

Also, because those buttons are hidden do we still need locales like this?

        no_subs = "No subtitles",
        no_audio = "No audio tracks",

Keith94:
Closes #587


Samillion:

Great catch with replacing tab with spaces, well done.

Merging.

PR #594, opened 2026-03-28
refactor: remove loop_in_pause option, keep behavior as default

Samillion:
**Changes**:
- Remove `loop_in_pause` option, keep behavior as default

I am not sure why this option was added, it's redundant.

Behavior unchanged:
- Right click on play/pause will toggle file loop
- Shift+Left click (or mid click) on play/pause will toggle playlist loop

Thanks @Xurdejl's for the suggestion


Keith94:
For some reason, osc.lua lets you customize it. However, removing the option definitely looks cleaner.

Samillion:
> For some reason, osc.lua lets you customize it. However, removing the option definitely looks cleaner.

True, we should allow customization on that as well I think. Will do in a different PR.

PR #597, opened 2026-03-28
feat: add play/pause mouse action options and optimize events

Samillion:
**Changes**:
- Add play/pause options for mouse actions
- Change Sub/Audio track buttons tooltip to be consistent with Playlist button
- Simplify various event responders logic

**Play/Pause options**:

```EditorConfig
# play/pause button mouse actions
play_pause_mbtn_mid_command=cycle-values loop-playlist inf no
play_pause_mbtn_right_command=cycle-values loop-file inf no
```

Thanks to @Keith94 for pointing out upstream mpv play/pause behavior.

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -256,8 +256,7 @@ Customize the button function based on mouse actions.
 |                               | sub_track_mbtn_right_command     | `cycle sub`                                  |
 |                               | sub_track_wheel_down_command     | `cycle sub`                                  |
 |                               | sub_track_wheel_up_command       | `cycle sub down`                             |
-| Play/Pause button             | play_pause_mbtn_left_command     | `cycle pause`                                |
-|                               | play_pause_mbtn_mid_command      | `cycle-values loop-playlist inf no`          |
+| Play/Pause button             | play_pause_mbtn_mid_command      | `cycle-values loop-playlist inf no`          |
 |                               | play_pause_mbtn_right_command    | `cycle-values loop-file inf no`              |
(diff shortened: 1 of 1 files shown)


nekoxuee:
Was removing the replay function intentional?

Samillion:
> Was removing the replay function intentional?

Yes. Hmmm, should I only make right and mid clicks customizable to keep the replay?

nekoxuee:
I ask because when eof is reached the icon changes to replay, but doesn't really work anymore since the seek to the start was removed

Samillion:
Good point. I'll add replay behavior back, it's useful, no sense in removing that.

nekoxuee:
Yeah, I was just wondering because the functionality is lost. I guess there is no native bind to do that right?

Samillion:
Done. Thank you for pointing it out.

Samillion:
I've been testing to make sure the adjustments I've made haven't caused any hiccups, but that doesn't give me confidence anymore.

I won't rush merging this PR, if anything is not working as it's supposed to, I would appreciate it very much to let me know.

Keith94:
>I guess there is no native bind to do that right?

In this case, you could theoretically apply some special options when `eof_reached` property is active.

```editorconfig
[End-of-file auto-profile]
profile-cond=eof_reached
profile-restore=copy-equal
script-opt=modernz-play_pause_mbtn_left_command=seek 0 absolute-percent; set pause no

script-opt=modernz-visibility=always
script-opt="modernz-osc_color=#804600"
```

It's too complicated for just a replay feature, but it's nice if you want to apply other things.

PR #600, opened 2026-03-28
fix: don't reset window or chapter title alpha styles

Samillion:
**Changes**:
- Don't reset window or chapter title alpha styles set in `osc_styles`
  - Add `lo.alpha[3] = 0` to all titles, to avoid default value `255`, which applies full transparency on alphas
- Don't use blur for titles, to give text a sharp look
- Adjust border for tooltip, window title, osc title and chapter title to be more pronounced (`0.5` => `1`)
- Rename `osc_styles.tooltip_seek` to `osc_styles.tooltip_box` for clarity
- Minor adjustments to cached hover effects use and `is_clickable` structure


Keith94:
Nice. Loving that new border look.

PR #604, opened 2026-03-29
feat: add semi transparent background hover effect

nekoxuee:
yet another hover effect, this time a box background

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -169,14 +169,14 @@ watch-later-options-remove=osd-margin-y
 
 ### Button hover settings
 
-| Option                   | Value           | Description                                                                                      |
-| ------------------------ | --------------- | ------------------------------------------------------------------------------------------------ |
-| hover_effect             | size,glow,color | active button hover effects: `"glow"`, `"size"`, `"color"`; can use multiple separated by commas |
-| button_hover_size        | 115             | relative size of a hovered button if "size" effect is active                                     |
-| button_glow_amount       | 5               | glow intensity when `"glow"` hover effect is active                                              |
-| slider_hover_effect      | yes             | apply size effect only when hovering slider handles                                              |
-| slider_hover_size        | 130             | relative size of a hovered slider handle if "slider_hover_effect" is used                        |
-| tooltip_hints            | yes             | enable tooltips for most buttons. seek and volume tooltips are always enabled                    |
+| Option              | Value               | Description                                                                                               |
+| ------------------- | ------------------- | --------------------------------------------------------------------------------------------------------- |
+| hover_effect        | size,glow,color,box | active button hover effects: `"glow"`, `"size"`, `"color"`, `"box"`; can use multiple separated by commas |
+| button_hover_size   | 115                 | relative size of a hovered button if "size" effect is active                                              |
+| button_glow_amount  | 5                   | glow intensity when `"glow"` hover effect is active                                                       |
+| slider_hover_effect | yes                 | apply size effect only when hovering slider handles                                                       |
+| slider_hover_size   | 130                 | relative size of a hovered slider handle if "slider_hover_effect" is used                                 |
+| tooltip_hints       | yes                 | enable tooltips for most buttons. seek and volume tooltips are always enabled                             |
 
 ### Progress bar settings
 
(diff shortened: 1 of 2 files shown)


Samillion:
I took it for a test drive. Even with all hover effects (`hover_effect=size,glow,color,box`), it looks great.

As we discussed on irc, the color of the hover box. Is it too light? Should we mimic a popular color palette? (ie: Youtube or something)

nekoxuee:
YouTube uses rgba(0, 0, 0, 0.3). Should I use that colour?

nekoxuee:
> YouTube uses rgba(0, 0, 0, 0.3). Should I use that colour?

nvm this is for background pill that contains YT buttons, when hovering is lighter

Keith94:
If I may chime in, does the color need to be static? What if the box follows `hover_effect_color` or otherwise default to a generic one.

Samillion:
> > YouTube uses rgba(0, 0, 0, 0.3). Should I use that colour?
> 
> nvm this is for background pill that contains YT buttons, when hovering is lighter

True. This is a tricky one, and honestly, I'm not sure if my comment about the color is valid.

However, the reason for my initial color change thoughts are:
- Bright, doesn't match the darker fades (at least in the gradiant scale)
- Glow makes the button look low quality in that bright color? Lower or no glow with the bright color improves it, but still feels off because it's too bright

As I said though, I'm not entirely sure if my initial thoughts are correct or no.

Nevertheless, excellent work.

I have to go for a few hours to pretend to be human in real life. I'll brainstorm during.

nekoxuee:
> If I may chime in, does the color need to be static? What if the box follows `hover_effect_color` or otherwise default to a generic one.

Yeah, I could do that. Originally skipped it to avoid the icon blending when both colour and box effects are enabled, but since the opacity now is only 10% should look fine.

<br>

> As I said though, I'm not entirely sure if my initial thoughts are correct or no.

Yes, they are correct, the opacity was too high. Now is 10% just like TY rgba(255, 255, 255, 0.1).

Keith94:
>Yeah, I could do that. Originally skipped it to avoid the icon blending when both colour and box effects are enabled, but since the opacity now is only 10% should look fine.

It does look very modern when it matches like that.

Samillion:
I'll check as soon as I'm back home.

By the way, I forgot to mention: Love the close button behavior, but it makes the X almost invisible. Maybe increase opacity? That way it can still appear reddish, without making the X invisible.

Keith94:
> I'll check as soon as I'm back home.
> 
> By the way, I forgot to mention: Love the close button behavior, but it makes the X almost invisible. Maybe increase opacity? That way it can still appear reddish, without making the X invisible.

Forcing the hover color to white seems to also solve this.

```lua
if hover_effect_box then
    user_opts.windowcontrols_close_hover = "#FFFFFF"
end
```

Samillion:
> Forcing the hover color to white seems to also solve this.

True! That would match Windows' behavior.

nekoxuee:
> Forcing the hover color to white seems to also solve this.

Done. Also, now the box background now uses hover_effect_color

Keith94:
The box background for bottom buttons could also fallback to `#FFFFFF` if user's `hover_effect` does not contain `color`, since the default `side/middle_buttons_color` is white.

Samillion:
Terribly sorry about the late response, ended up spending the night at a friend's.

Looks absolutely beautiful, well done. Everything is perfect except I feel the close (X) button is like a panic button.

The color effect on minimize and maximize is very subtle and nice, maybe it should just share the same behavior instead of a custom one?

Or am I overthinking again? lol

Again, terribly sorry about the delay.

Samillion:
Beautiful, really. You're not going to enable box effect by default? It looks really good.

I want to coordinate merging this and @Keith94's tooltip PR, they work well together.

nekoxuee:
Do I enable all by default?

Samillion:
> Do I enable all by default?

You don't think we should? It looks really good. I'm ok with what you think is right by default, just feels like a shame most users won't be aware that this exists.

P.S. I'd love your input here, if it's not a bother: https://github.com/Samillion/ModernZ/pull/606#issuecomment-4155548826

Samillion:
Thanks again your efforts. Merging.

PR #606, opened 2026-03-30
feat: anchor tooltips above seekbar with styled pill box

Keith94:
**Changes:**

* Button and volumebar hover tooltips now appear at a consistent position
* Extend pill box to volumebar and button tooltips
* Increase horizontal padding back to 5px
* Remove redundant options

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -221,7 +221,7 @@ Useful when adjusting font size or type, this will help you change the affected 
 | tooltip_height_offset       | 5     | tooltip height position offset                                         |
 | portrait_window_trigger     | 1000  | portrait window width trigger to move some elements                    |
 | hide_volume_bar_trigger     | 1150  | hide volume bar trigger window width                                   |
-| osc_height                  | 70    | osc height                                                             |
+| osc_height                  | 60    | osc height                                                             |
 
 ### Mouse Commands (User Options)
 
--- a/modernz.conf
+++ b/modernz.conf
@@ -331,14 +331,14 @@ title_with_chapter_offset=4
 chapter_title_offset=20
 # time codes vertical offset relative to seekbar
 time_codes_offset=0
-# time codes vertical offset with portrait window
+# tooltip height position offset
 tooltip_height_offset=5
 # portrait window width trigger to move some elements
 portrait_window_trigger=1000
 # hide volume bar trigger window width
 hide_volume_bar_trigger=1150
 # osc height
-osc_height=70
+osc_height=60
 
 ## Mouse commands
 ## details: https://github.com/Samillion/ModernZ#mouse-commands-user-options
--- a/modernz.lua
+++ b/modernz.lua
@@ -206,7 +206,7 @@ local user_opts = {
     tooltip_height_offset = 5,             -- tooltip height position offset
     portrait_window_trigger = 1000,        -- portrait window width trigger to move some elements
     hide_volume_bar_trigger = 1150,        -- hide volume bar trigger window width
-    osc_height = 70,                       -- osc height
+    osc_height = 60,                       -- osc height
 
     -- Mouse commands
     -- customize the button function based on mouse action
@@ -1456,7 +1456,7 @@ local function render_elements(master_ass, osc_vis, wc_vis)
                             local seekbar = state.seekbar_element
                             local ref_el = (element.name == "volumebar" and seekbar and seekbar.hitbox) and seekbar or element
                             local image_mode_offset = (ref_el == element) and 10 or 0
-                            ty = ref_el.hitbox.y1 - user_opts.tooltip_height_offset - image_mode_offset
+                            ty = ref_el.hitbox.y1 + elem_geo.h / 2 - user_opts.tooltip_height_offset - image_mode_offset
                         else
                             ty = element.hitbox.y1 + elem_geo.h / 2 - user_opts.tooltip_height_offset
                         end


Samillion:
Terribly sorry about the late response, ended up spending the night at a friend's.

Excellent work. One issue, if you have `osd-border-style=background-box` in your `mpv.conf`, it'll definitely draw a secondary box inside the tooltip.

The unified function you made is great, but since it does two things, you need to reset to default state twice. (for each event)

For example:

```lua
-- draw a pill-shaped tooltip background box and label
local function draw_tooltip(ass, tx, ty, width, style, label, alpha)
    local fs = user_opts.tooltip_font_size
    local ph, pv = 5, 3
    -- draw tooltip box
    ass:new_event()
    ass:append("{\\rDefault\\alpha&H80&}")
    ass:pos(tx - width / 2 - ph, ty - fs - pv)
    ass:an(7)
    ass:append(osc_styles.tooltip_box)
    ass:draw_start()
    ass:round_rect_cw(0, 0, width + 2 * ph, fs + 2 * pv, (fs + 2 * pv) / 2)
    ass:draw_stop()
    -- add tooltip
    ass:new_event()
    ass:append("{\\rDefault}")
    ass:pos(tx, ty)
    ass:an(2)
    ass:append(style)
    if alpha then ass_append_alpha(ass, alpha, 0) end
    ass:append(label)
end
```

I'll be able to help better once I'm back home. Again, terribly sorry about the delay.

Keith94:
Ah, this is why I rely on your expertise lol. Glad you like it though.

Please investigate, make any changes you need or  I can adjust that later. 🙏

Samillion:
Looks great. I have a question, but keep in my mind these are questions, not suggestions.

- Should the tooltips show above seek bar? In a way, they are unified, but does it look right?
- Would making them show just above the button be a bad idea?
  - We can adjust box transparency so that it would be clear to read, even if it's rendered on top of the seek bar
  - For layering, we can lower seek bar to layer `49`, so tooltips always render above it
    - Then we can layer title and chapter title at `48` so that thumbnail/time stamp seek tool tip always render above them

Keith94:
>Would making them show just above the button be a bad idea?

I initially felt it looked too busy with tooltips and multiple elements on top of each other, especially with the new box effect.

Since we can now change osc height,  I was going to suggest decrease the height to 62-63 for a more minimal look once the tooltips were out of the way. YouTube has a similar approach.

But let me know what you think.

Samillion:
> Since we can now change osc height, I was going to suggest decrease the height to 62-63 for a more minimal look once the tooltips were out of the way. YouTube has a similar approach.

This sounds very nice. Agreed.

nekoxuee:
Since this is inspired by YouTube:
The area above YT progress bar is empty and the gap between the buttons and the progress bar is too tight, that is why they decided to place the tooltip above the progress bar.

Samillion:
If we do decide OSC height to be 62 (or 60?), we should:
- Fix layering so tooltips render above Title/Chapter
- Adjust button tooltip transparency to not be so transparent, for readability.
  - By default, should we disable title/chapter title above osc? Question, not a suggestion

Samillion:
Sorry for the additional comment.

Right now, the seek bar tooltips (time/chapter/thumbnail) are drawn on-top of title/chapter title. So the behavior already exists, right? So having all tooltips show there, unified, wouldn't be a problem, right?

So it all goes to the questions from my previous comment

> If we do decide OSC height to be 62 (or 60?), we should:
> - Fix layering so tooltips render above Title/Chapter
> - Adjust button tooltip transparency to not be so transparent, for readability.
>   - By default, should we disable title/chapter title above osc? Question, not a suggestion

nekoxuee:
I think the title and chapter is basic info and shouldn't be disabled by default.

Samillion:
Alright. What about the other adjustments, good?

Or do you think button tooltips should not be above seek bar?

Samillion:
Sorry, correction:
- Fix layering so tooltips render ~~above~~ on top of Title/Chapter (like thumbnail)

nekoxuee:
> Or do you think button tooltips should not be above seek bar?

I don't really mind the tooltips being above the progress. I just pointed out the reason why YT does it this way is due to their constraints

Most of the time I don't really know what to say because many changes feel like opinions and there isn't a consistent rule to follow, plus I don't use modernz myself, so I don't really know what users expect or prefer. I usually just leave these kinds of decisions up to you, either way is fine.

Samillion:
Transparency wise, what about: ~30%

```lua
ass:append("{\\rDefault\\alpha&H4D&}")
```

image


To help:

```
100%  - &HFF&  (invisible)
 90%  - &HE6&
 80%  - &HCC&
 70%  - &HB3&
 60%  - &H99&
 50%  - &H80&
 40%  - &H66&
 30%  - &H4D&
 20%  - &H33&
 10%  - &H1A&
  0%  - &H00&
```

Samillion:
**Changes**:
- Adjust layers so tooltips does not render behind title/chapter
  - Change seek bar layer to `49` instead of `51`
  - Change title and chapter title layer to `48` instead of `50` (to render behind thumbnail)
- Adjust tooltip box transparency to `40%`

image

<br>

image

Keith94:
30% alpha is a lot easier to read especially when it's overlapping other text.

A height of 60 looks great as well.

Samillion:
**Changes**:
- Fix `hitbox` crash in image mode
- Fix image mode tooltip offset
- Adjust tooltip transparency to `30%` instead of `40%`

Samillion:
Should we adjust default value of `osc_height` to `60`? I honestly like it, but I'm not sure if it's personal preference or actually a good change.

Also, are we sure alpha `30%` is not too opaque? or was `40%` unreadable? This was `40%`:

image

Keith94:
I can confirm when I'm home. I'll spend more time with it. Thanks for all the fixes btw

Samillion:
I now see how 30% is better (what is currently used), definitely makes reading tooltip consistent, especially for old farts like me who might squint to read stuff. 😝 

I didn't have problems with 40% though, so I'm fine with either one.

Keith94:
The seek tooltip appears to be anchored a little higher than button tooltips. It should probably come down to match.

>Should we adjust default value of osc_height to 60? I honestly like it, but I'm not sure if it's personal preference or actually a good change.

I would also agree to set to 60; there's still a good amount of padding so nothing overlaps.

Samillion:
Sounds good. Could you make the adjustments please? I'm currently preparing a locale change, if you don't mind and if it's not a bother.

Keith94:
No worries, I made the changes.

Samillion:
Beautiful, thank you very much. Ready to merge? I've tested:
- `modern` layout
- `modern-compact` layout
- `modern-image` layout (needs some fixing, but it's prepared already)
- with and without thumbfast
- with @Xurdejl's box hover

image

So far, everything is looking very noice.

Keith94:
>Beautiful, thank you very much. Ready to merge?

Looks good to me! Great job all.

Samillion:

Keith94:
You as well.

It's oddly satisfying seeing all the repetitive code nuked out of existence.

Samillion:
> It's oddly satisfying seeing all the repetitive code nuked out of existence.

Indeed. All thanks to the function you created, well done.

Side note: My god, updating all the previews was a pain in the ass lol.

I want a better main preview image, with a higher quality, but I can't be bothered after all that. 😝

Keith94:
Previews look very nice. Never even thought to set options through console like that.

Not sure what happened, but the volume slider tooltip is offset slightly lower. Weird.

Samillion:
> Not sure what happened, but the volume slider tooltip is offset slightly lower. Weird.

Ah, nice catch. It's the added `+ elem_geo.h / 2` in https://github.com/Samillion/ModernZ/pull/606/changes/b50e29955e7cd6f157ec1a13b26b55c514dab2f9

Could be solved by setting an offset

```diff
-ty = ref_el.hitbox.y1 + elem_geo.h / 2 - user_opts.tooltip_height_offset - image_mode_offset
+local offset = (ref_el == element) and (elem_geo.h / 2) or 0
+ty = ref_el.hitbox.y1 + offset - user_opts.tooltip_height_offset - image_mode_offset
```

But I don't like using the name `local offset`, too common. `point_offset`? `middle_offset`? `anchor_offset`?

I dunno, I hate naming things lol.

Keith94:
That does indeed solve it. Do you want me to add that?

Samillion:
> That does indeed solve it. Do you want me to add that?

Yes please, that would be awesome. I'm working on optimizing loading external locales at the moment.

Keith94:
You are a machine.. lol

I'll try to combine it with another fix (hopefully I won't break anything else).

Samillion:
> You are a machine.. lol

😆 The code base is so close to actually reaching a clean and well optimized state. I cannot wait for that to happen, because it means less frequent updates and a stable behavior overall.

> I'll try to combine it with another fix (hopefully I won't break anything else).

Thank you very much ❤️

Issue #251, opened 2024-11-25
Show window controls when idle


Keith94:
### Expected behavior of the wanted feature

Enable `window_top_bar` functionality when mpv is on idle screen (no file loaded).

Workaround for enabling **native** title bar only on idle screen: https://github.com/Samillion/ModernZ/issues/251#issuecomment-4019194253

Samillion:
Oh wow, I didn't even realize this doesn't show, even on stock osc! I wonder why.

Keith94:
> Oh wow, I didn't even realize this doesn't show, even on stock osc! I wonder why.

Yeah it's strange indeed; there's a feature request for it here: https://github.com/mpv-player/mpv/issues/13167

Anthonyy232:
This would actually be awesome.

zenwav:
I second this.

Samillion:
I'm still trying to figure out why mpv hasn't implemented this by default in `osc.lua`. I mentioned it a couple of times on irc, but I wasn't always present the last month.

Keith94:
True. Would be simple to port it over once they add it officially.

Samillion:
Indeed. One of the main things I wanted to keep with ModernZ is that it's based on `osc.lua`.

This specific feature would require a drastic change (osc render). Unless there is a simple way that my brain is unable to think of yet.

Keith94:
There's another popular 3rd party OSC that implemented it, but I didn't look into how he did it.

Keith94:
I found some property called idle-active but no idea if that's even helpful.

`mp.observe_property('idle-active', 'bool', function(_, idle)`

[source](https://github.com/tomasklaen/uosc/commit/b8a37ef576cf51b33ab859012245cc787cefe536)

Samillion:
> There's another popular 3rd party OSC that implemented it, but I didn't look into how he did it.

What's their repo link? I'd love to see their implementation.

Keith94:
https://github.com/tomasklaen/uosc

Samillion:
Ohh. Yeah. To be very frank, uosc is an amazing implementation as an osc overall that broke the osc standard and went over and beyond. It is the closest thing to a GUI toolkit yet it's just an OSC.

To be clear, this feature is doable. A simple (broken) example, in the tick function:
https://github.com/Samillion/ModernZ/blob/80ea6b715412ea339ad8097a4e8c7cc014d344a7/modernz.lua#L3112-L3117

It controls osc render:
[...]

PR #615, opened 2026-03-31
feat: show window controls in idle

Samillion:
Fixes: https://github.com/Samillion/ModernZ/issues/251 (Finally!)

**Changes**:
- Show window controls in idle (on hover)
  - Based on upstream mpv
  - In idle, zone hover mode is always set to `independent`
  - Zone hover mode is restored to what's used in `user_opts` when not in idle
  - Add `saved_hover_mode_idle` state, saves `zones_hover_mode` when idle overrides it to `independent`
- Match helpers with upstream mpv: `set_osd()` and `render_wipe()`
- Remove redundant `window_title` option, it will now respect `--title` set by user
  - They're both window bar titles, no sense in having to set it twice

image

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -564,7 +564,6 @@ local state = {
     hide_timer = nil,
     demuxer_cache_state = nil,
     idle_active = false,
-    saved_hover_mode_idle = nil,            -- saved zones_hover_mode when idle overrides it to "independent"
     audio_track_count = 0,
     sub_track_count = 0,
     playlist_count = 0,
@@ -3286,7 +3285,7 @@ local function mouse_leave()
         if not state.pause_osc_locked then
             hide_osc()
         end
-        if user_opts.zones_hover_mode == "independent" then
+        if user_opts.zones_hover_mode == "independent" and not state.idle_active then
             hide_wc()
         end
     end
@@ -3511,7 +3510,7 @@ local function render()
 
     --mouse input area
     local wc_vis
-    if user_opts.zones_hover_mode == "independent" then
+    if user_opts.zones_hover_mode == "independent" or state.idle_active then
         wc_vis = state.wc_visible
     else
         wc_vis = state.osc_visible
@@ -3571,7 +3570,9 @@ local function render()
 
     if state.hide_timer then state.hide_timer.timeout = math.huge end
     run_autohide("showtime",    hide_osc, osc_areas)
-    run_autohide("wc_showtime", hide_wc,  wc_areas)
+    if not state.idle_active then
+        run_autohide("wc_showtime", hide_wc,  wc_areas)
+    end
 
     -- actual rendering
     local ass = assdraw.ass_new()
@@ -3635,6 +3636,11 @@ tick = function()
         end
 
         if window_controls_enabled() then
+            -- ensure window controls are visible during idle for all modes
+            -- use show_wc() so wc_showtime is set and autohide doesn't immediately fire
+            if not state.wc_visible then
+                show_wc()
+            end
             render()
         else
             render_wipe(state.osd)
@@ -3748,22 +3754,7 @@ end)
 observe_cached("border", request_init_resize)
 observe_cached("title-bar", request_init_resize)
 observe_cached("window-maximized", request_init_resize)
-observe_cached("idle-active", function()
-    if state.idle_active then
-        -- idle: apply override
-        if user_opts.zones_hover_mode ~= "independent" then
-            state.saved_hover_mode_idle = user_opts.zones_hover_mode
-            user_opts.zones_hover_mode = "independent"
-        end
-    else
-        -- not idle: restore hover mode
-        if state.saved_hover_mode_idle then
-            user_opts.zones_hover_mode = state.saved_hover_mode_idle
-            state.saved_hover_mode_idle = nil
-        end
-    end
-    request_tick()
-end)
+observe_cached("idle-active", request_tick)
 mp.observe_property("user-data/mpv/console/open", "bool", function(_, val)
     if val and user_opts.visibility == "auto" and not user_opts.showonselect then
         osc_visible(false)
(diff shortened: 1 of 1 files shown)


nekoxuee:
Can't seem to get them to show. I tried with the osc.lua file from mpv's pr and it works. Is there something I'm missing?

Samillion:
> Can't seem to get them to show. I tried with the osc.lua file from mpv's pr and it works. Is there something I'm missing?

Only thing that's needed is to disable `title-bar` or `border` in `mpv.conf`, then hover top bar to show based on your hover zone mode

nekoxuee:
I tried both border=no and title-bar=no, and they don't show. The weird thing is that I drop osc.lua without changing anything and that works on idle showing the buttons

Samillion:
> I tried both border=no and title-bar=no, and they don't show. The weird thing is that I drop osc.lua without changing anything and that works on idle showing the buttons

How... it's the same exact implementation. Any errors?

Why is it working for me 😞 @Keith94 is it working for you?

Keith94:
It works for me with border=no or title-bar=no

nekoxuee:
I don't understand. I have no config. Just osc=no and title-bar=no.

Samillion:
> I don't understand. I have no config. Just osc=no and title-bar=no.

That is so weird. I tried with different combinations and it still works. (ie: `show_window_title=no`)

Any indication or error when you run in terminal?

```shell
mpv --idle --force-window --v
```

https://github.com/user-attachments/assets/2ca65278-fa85-44ca-a694-7f156ea2eaf3

Samillion:
Wait, I think I found out why, with an empty `modernz.conf` it doesn't show.

One of the default options... hmmmm

nekoxuee:
> Wait, I think I found out why, with an empty `modernz.conf` it doesn't show.
> 
> One of the default options... hmmmm

If it's a user option needs to be related to showing the osc on pause, since stock osc.lua does not have that logic

Samillion:
It's `zones_hover_mode`.
- `always`: does not work (default)
- `zones`: does not work
- `independent`: works

Samillion:
This is pissing me off. I'm almost tempted to just make the whole hover zones thing changed to osc.lua's deadzonesize behavior.

Samillion:
Actually, would that be a bad idea?

Removing `zones_hover_mode` option completely, including related options: `bottomhover_zone`, `tophover_zone`

Then change the whole behavior to match osc.lua's `deadzonesize` option.

Keith94:
If the motivation for the change is for idle compatibility then I can understand that.

If it were up to me, I'd keep it modernz-style: instead of hiding osc immediately when cursor enters deadzone, respect hidetimeout value always. If that's possible of course.

Samillion:
It would be a huge refactor. I'll try to fix it as is first, before doing anything drastic. 

**Refactor**:
- Removing `zones_hover_mode` option completely, including related options: `bottomhover_zone`, `tophover_zone`
- Adjusting `render_elements()`, `show_osc()`, `hide_osc()`, `mouse_leave()`, `reset_timeout()`, `process_event()`, `enable_osc()`, `render()`, `always_on()`, `mp.observe_property("pause")`, `keeponpause`, `showonpause`, and `validate_user_opts()`
- Add deadzonesize option and related helpers

Such a huge change will definitely introduce a whole new set of bugs.

Keith94:
If that's the case I have complete faith in you. 😁

nekoxuee:
If that's easier to maintain then go for it.

Samillion:
It won't be in this PR to be honest.

I'm thinking about just making `zones_hover_mode=independent` in idle, otherwise (not idle) go with how it's set in `user_opts`.

The `deadzonesize` refactor is massive,, I can experiment with it soon, but I can't promise anything lol.

Sorry for not applying a fix yet, I'm taking a brain recovery break for a day or so.

Samillion:
**Changes**:
- Show window controls in idle with all zone hover modes
  - In idle, zone hover mode is always set to `independent`
  - Zone hover mode is restored to what's used in `user_opts` when not in idle
  - Add `saved_hover_mode_idle` state, saves `zones_hover_mode` when idle overrides it to `independent`

Samillion:
I'll experiment with deadzonesize soon, but honestly just looking at all the changes is making my brain explode.

Can't AI handle this crap? "Yo, port over osc.lua's stuff, only the ones we need without breaking stuff"

Keith94:
What do you think about this one?
https://github.com/Keith94/ModernZ/blob/idle_testing/modernz.lua

The WC seem to show permanently in idle (in all modes). I kinda like the result.

Samillion:
> What do you think about this one? https://github.com/Keith94/ModernZ/blob/idle_testing/modernz.lua
> 
> The WC seem to show permanently in idle (in all modes). I kinda like the result.

This part:

```lua
            -- ensure window controls are visible during idle for all modes
            -- use show_wc() so wc_showtime is set and autohide doesn't immediately fire
            if not state.wc_visible then
                show_wc()
            end
```

It will call `show_wc()` per tick. So (by default) about 8 times per second. or am I misunderstanding it?

Samillion:
Actually nevermind, it works well, no OSD cache build up.

Keith94:
AFAIK it fires once per idle entry; once show_wc() succeeds and sets state.wc_visible = true, it becomes a no-op.

Samillion:
You are absolutely right, thank you very much.

Samillion:
Well, crap. Yeah, it doesn't run it per tick, but per mouse move.

So for example, in `zones_hover_mode=always`, it runs per pixel.

How to test:
- Set `zones_hover_mode=always` in `modernz.conf`
- Run `mpv --idle --force-window --v`
- Move mouse inside window
  - Margin spam per pixel move

Samillion:
Wth, I reverted but it's not applying it to the PR.

https://github.com/Samillion/ModernZ/commit/23a08f1040d8958eb28764c12ee47d3adb8fd913

PR #618, opened 2026-04-01
refactor: window controls keep hover color when held

Keith94:
Window controls now retain their hover color when clicked and held, instead of switching to `held_element_color`.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -1410,7 +1410,7 @@ local function render_elements(master_ass, osc_vis, wc_vis)
             end
             if mouse_hit(element) then
                 -- mouse down styling
-                if element.styledown then style_ass:append(osc_styles.element_down) end
+                if element.styledown then style_ass:append(element.hover_color and ("{\\1c&H" .. osc_color_convert(element.hover_color) .. "&}") or osc_styles.element_down) end
                 if element.softrepeat and state.mouse_down_counter >= 15 and state.mouse_down_counter % 5 == 0 then
                     element.eventresponder[state.active_event_source.."_down"](element)
                 end


Samillion:
Isn't this intended behavior though? To differentiate from what is pressed and what isn't?

Keith94:
In stock osc, it looks fine there because buttons are white and held color is gray.

In modernz, the default hover colors are green, yellow, red, so idk if using gray/custom held color seems out of place in this context.

The PR could be expanded to explore alternative ideas for pressed buttons, though. Either decreasing size by 5% or changing hover bg transparency etc.

Samillion:
I apologize about the delay in response, `deadzonesize` was taking all my attention.

> In modernz, the default hover colors are green, yellow, red, so idk if using gray/custom held color seems out of place in this context.

If you press/hold down, hover away then hover back, it still shows a clear indicator that it is the one being pressed.

With this change, no indicator whatsoever.

If the issue is the gray color, maybe another color variant instead?

Objectively speaking, the way this PR alters the behavior reduces possible visual information of what is happening, which will be seen as a downgrade in function to match cosmetics in a niche scenario.

Sorry, I genuinely don't mean to reject, I'm merely giving you my honest thoughts, and I could be wrong.

Keith94:
I get what you mean. I don't think there's a held color that fits well with all 3 buttons unless you add separate colors which seems overkill.

Or an idea is make the hover background a bit brighter when button is held. Pretty common in a lot of UIs.

Samillion:
Terribly sorry for not responding, I wasn't ignoring, I was distracted by seekbar height/keep with osc PR.

Regardless, like I said, this was not a rejection, just a brainstorm of thoughts. If you can think of a way to improve aesthetics without function compromise, by all means go for it.

Keith94:
Nono you're absolutely right. I need a better approach lol. Will think this one over a bit,

PR #627, opened 2026-04-03
fix: keep consistent handle size during drag

Keith94:
Applies `slider_hover_effect` size increase during seekbar handle drag. This avoids the issue where size shrinks back down if the handle position lags behind the cursor during drag.


Samillion:
Damn, nice catch. Thank you very much.

PR #630, opened 2026-04-03
feat: add `gap` nibble style

Keith94:
Adds a new `nibbles_style = "gap"` option that replaces the seekbar background with a segmented bar divided by transparent gaps at each chapter boundary.

The first chapter marker is no longer drawn as a nibble in any style. In most files the first chapter starts at position 0; placing a nibble at the very beginning of the seekbar serves no useful purpose.

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -155,6 +155,8 @@ watch-later-options-remove=osd-margin-y
 | hover_effect_color                | `#FB8C00` | color of a hovered button when `hover_effect` includes `"color"`                                  |
 | thumbnail_box_color               | `#111111` | color of the background for thumbnail box                                                         |
 | thumbnail_box_outline             | `#404040` | color of the border outline for thumbnail box                                                     |
+| nibble_color                      | `#FB8C00` | color of chapter nibbles on the seekbar                                                           |
+| nibble_current_color              | `#FFFFFF` | color of the current chapter nibble on the seekbar                                                |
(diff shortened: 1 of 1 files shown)


Samillion:
Awesome, thank you very much. I'll be home soon to test thoroughly, very exciting stuff.

Sorry if my brain doesn't work. Assuming that `sr` and `sl` are segment right and left, this can probably be optimized by adding a helper that handles gap cuts. ie: in `draw_seekbar_ranges()`, `draw_seekbar_nibbles()` and `draw_seekbar_progress()`

Hmmm, but then `draw_gap_segments()` will need exclusions...

My brain fried lol.

Either way, well done. Can't wait to see it in action.

Edit: If the gaps were like the screenshots in the discussion, can they be decreased by like half a pixel? or even less than half, just a bit thinner. or is that a bad idea?

Keith94:
>Edit: If the gaps were like the screenshots in the discussion, can they be decreased by like half a pixel?

Yeah, I reduced it by 1px. At 1.1 scale it looks like this.

msedge_9Fw50PzfVw

Samillion:
I take back what I said, no matter how much I change it or try to optimize, in the end it's a just a restructure, not an actual optimization.

I honestly got lost in the code a few times lol, 

Anywho, the code is solid. Any editing or changing would be... I don't know, for code readability. At least that's what I could come up with.

Let me know when it's merge ready, please.

Keith94:
All seems good on my end. I've tested with different height and handle sizes, scales, seekrange etc.

Make any changes you'd like and it's good to merge.

PR #631, opened 2026-04-03
feat: add size and box alpha for held buttons

Keith94:
**Added:**

- `button_held_size` - sets the scale of a button when held/pressed, independent of button_hover_size. Previously, holding a button always reset the scale back to 100%.

- `button_held_box_alpha` - sets the transparency of the hover background box when a button is held down. Defaults to 0xCC (~80% transparent), brighter than the standard hover box at 0xE6 (~90% transparent).


Samillion:
Thank you very much.

My only note is the way alpha will be set by users, at least 90% of users will not know what that value means or what it should be.

Would my opacity convert function from pause indicator help you here? so that it would be a 0-100 value instead.

https://github.com/Samillion/ModernZ/blob/79fbbfb5a0085b841025434b48e3d351ce968a63/extras/pause-indicator-lite/pause_indicator_lite.lua#L95-L99

Keith94:
That's a great idea. I've added it to the PR now.

Samillion:
Awesome, thank you very much. Let me know when it's merge ready, please.

**Side note**:
This PR and the gap chapter PR made me realize there is a whole bunch of dead code, wrong code and redundant code still left.

Specifically in `osc_styles`, `layouts` and `osc_init`.

image

Keith94:
It's good to go.

Still it's an accomplishment how much code you got cleaned up. I'll try to help anyways.

Issue #628, opened 2026-04-03
speed_button=no causes overlap in narrow window


Keith94:
### mpv Information

```bash

```

### Details & Reproduction Steps

1. speed_button=no
2. Play a vertical video

Left side (speed_button=no) | Right side (speed_button=yes)

<img width="1099" height="181" alt="Image" src="https://github.com/user-attachments/assets/92674a07-2fcc-4796-873b-48e6710267d0" />

### I carefully read all instructions and confirm that I did the following:

- [x] I tested with the latest mpv version to validate that the issue is not already fixed.
- [x] I provided all required information.

PR #633, opened 2026-04-04
fix: don't tie osc title with buttons visibility

Samillion:
Fixes: https://github.com/Samillion/ModernZ/issues/628

**Changes**:
- Don't tie osc title with buttons visibility
- Improve `modern` layout elements visibility
  - Move all visibility conditions related to `modern` layout, inside `modern` layout, so that later once it's done for all layouts, they're removed from `osc_init()`


Samillion:
This PR is phase 1.

Next phase will be `modern-image` layout.

Full plan:
- `modern` layout cleanup
- `modern-image` layout cleanup
- `modern-compact` layout cleanup
- `osc_init()` cleanup

PR #637, opened 2026-04-04
fix: pause playback on seek handle click

Keith94:
Previously, when `mouse_seek_pause=yes` the video would not pause when clicking seek bar/handle until mouse is moved.

Now video pauses on mbtn_left_down (not wait for mouse_move).


Samillion:
Perfect. Thank you very much for your effort.

Merging.

PR #639, opened 2026-04-05
refactor: unify volume tooltip and hitbox

Keith94:
The volume elements now share the same tooltip and hover background appearance when mouse is over them.


Samillion:
That is brilliant, well done. Absolutely love the behavior now.

Merging.

PR #626, opened 2026-04-03
feat: add option to show chapter above title

clivedc:
adds the option `chapter_above_title`. Default is set to `no`

`chapter_above_title=no`
Screenshot_03-04-26_06:51:03PM

`chapter_above_title=yes`
Screenshot_03-04-26_06:51:43PM

PR #641, opened 2026-04-05
fix: factor in time codes offset with title to remain clickable

Samillion:
Fixes: https://github.com/Samillion/ModernZ/pull/626#issuecomment-4187349164

**Changes**:
When `show_title=no` and `show_chapter_title=yes`, `title_and_chapter_h_with_offset` affects `osc_height`. Now it factors in time codes offset to always have a correct hitbox above seekbar in both `modern` and `modern-compact` layouts.


Keith94:
Verified fixed 😍

PR #642, opened 2026-04-05
fix: only update visibility mode if it actually changed

Keith94:
Prevents `visibility_mode` from being needlessly re-evaluated (and disrupting state) every time an unrelated option like `shuffle_button` changes via autoprofile.


Samillion:
Beautiful, great idea. Thank you very much.

However, I promised @clivedc that I will not merge anything until his PR is done lol, I've done it to him before already and I feel guilty.

This PR will definitely not affect his PR, but still. He deserves that at least. 

I shall merge very soon.

PR #665, opened 2026-04-11
fix: suppress button tooltips when force_seek_tooltip is active

Keith94:
Prevents the issue of the tooltips overlapping when hovering a button while seeking and using `force_seek_tooltip`.


Samillion:
Awesome. Thank you very much.

Merging.

PR #666, opened 2026-04-11
pip-lite: improve mode restore to cover multiple scenarios

Samillion:
**Changes**:
Improve overall logic when restoring from PiP to normal, so it covers different scenarios: was maximized, fullscreen, windowed.

- Add more states in `state = {}` list to be more accurate on restore
- Add helpers to reduce code repetition: `cancel_geometry_timer()`, `capture_state()`
- Improve logic behavior going back to, or from PiP
  - Save original state (maximized, fullscreen, windowed)
  - Add a delay when state is changing to avoid any condition racing
  - Save original `autofit` options (if they were set in `mpv.conf`), to restore them accurately

### DIFF
--- a/extras/pip-lite/pip_lite.lua
+++ b/extras/pip-lite/pip_lite.lua
@@ -2,8 +2,8 @@
 
     This script adds a PiP mode to ontop (pin).
     When ontop is enabled, window is resized and moved to bottom right.
-    Fit and position can by adjusted in options.
-    
+    Fit and position can be adjusted in options.
+
     Tested on Windows only.
 
     More info: https://github.com/Samillion/ModernZ/tree/main/extras/pip-lite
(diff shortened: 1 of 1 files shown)


Keith94:
Exit pip goes back to fullscreen now. I was really surprised. Nice change.

Samillion:
The one thing I'm afraid to touch is restoring window position based on where it was and not a preset center/option.

I researched it a bit, and the difference (in Linux especially) is a nightmare to understand. X11, Wayland... whatever compositors and window managers...etc

I don't claim to understand any Linux related stuff, I'm just saying when I saw it, I got scared so I was like "nope, good enough like dis". 😆

Keith94:
Agreed, at least it works uniformly across platforms

PR #673, opened 2026-04-16
refactor: cleanup redundant osc_styles tags

nekoxuee:
- Remove explicit \blur0 and \bord0 as they are default ASS values.
- Strip border colour tags where border thickness is 0

Second commit is just a lsp linting error.


Samillion:
Awesome, thank you very much.

Merging.

PR #675, opened 2026-04-16
fix: adjust offset based on time codes position

Keith94:
In the modern layout when `title_and_chapter_h_with_offset == 0` and time codes were below seekbar, the offset was too tall and prevented mouse clicks above the seekbar. Now it's adjusted based on time codes being above or below.


Samillion:
Thank you very much.

Merging.

PR #676, opened 2026-04-17
pause-indicator: add option to hide indicator when dragging seekbar

Keith94:
Adds an option to control whether the pause indicator shows up while seeking and `mouse_seek_pause=yes`.


Samillion:

However, this is all just a complex way to make the indicator not show a correct pause state even though the media is paused.

PR #677, opened 2026-04-17
refactor: don't apply held size to text elements

Keith94:
Similar to hover size, text elements are no longer affected when `button_held_size=` is set to something other than 100.


Samillion:
Thank you very much.

Merging.

PR #678, opened 2026-04-17
refactor: expand hitbox for zoom and volume sliders

Keith94:
Expand hitbox vertically for zoom and volume sliders so they are easier to interact with.


Samillion:
Thank you very much.

Merging.

PR #680, opened 2026-04-17
feat: hide audio and subttiles track button if there is only one track

bennyyip:
https://github.com/bennyyip/ModernZ/commit/5cf800d9211876a0259b9dbf93f7c2dbd98e2a2d#diff-d5f739bcfd17a07337246b9195b4fd31e4624b22622aaf2e8e8265f4820789b1R2352-R2363
I am not sure if `time_codes_x` is correct.

The option names `hide_subtitles_button_if_one` and `hide_audio_tracks_button_if_one` is not so good. Do you have better idea?

### DIFF
--- a/modernz.conf
+++ b/modernz.conf
@@ -157,6 +157,10 @@ playlist_button=yes
 hide_empty_playlist_button=no
 # gray out the playlist button when no playlist exists
 gray_empty_playlist_button=no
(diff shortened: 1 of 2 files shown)


Samillion:
Thank you very much for your effort. However, this has been discussed before and the conclusion was that one track is still informative details on what's currently playing at a glance.

PR #682, opened 2026-04-18
feat: add play_pause left click command

Keith94:
Makes left click customizable during regular playback to match upstream mpv.

This PR would allow adding extra things on user's side; for example a script message to show a pause indicator:

```editorconfig
play_pause_mbtn_left_command=script-message toggle-pause-indicator; cycle pause
```
For more context, in my setup I try to avoid showing an indicator when seeking or frame stepping etc so I rely on this script message to show it in specific ways.

### DIFF
--- a/docs/USER_OPTS.md
+++ b/docs/USER_OPTS.md
@@ -258,7 +258,8 @@ Customize the button function based on mouse actions.
 |                               | sub_track_mbtn_right_command     | `cycle sub`                                  |
 |                               | sub_track_wheel_down_command     | `cycle sub`                                  |
 |                               | sub_track_wheel_up_command       | `cycle sub down`                             |
-| Play/Pause button             | play_pause_mbtn_mid_command      | `cycle-values loop-playlist inf no`          |
+| Play/Pause button             | play_pause_mbtn_left_command     | `cycle pause`                                |
+|                               | play_pause_mbtn_mid_command      | `cycle-values loop-playlist inf no`          |
 |                               | play_pause_mbtn_right_command    | `cycle-values loop-file inf no`              |
 | Chapter skip buttons          | chapter_prev_mbtn_left_command   | `add chapter -1`                             |
 |                               | chapter_prev_mbtn_mid_command    | `show-text ${chapter-list} 3000`             |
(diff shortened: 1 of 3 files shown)


Samillion:
I'm concerned about one scenario conflict.

1. Click (press down) `play_pause_mbtn_left_command=cycle pause`
2. Unhold click (press up) `mp.command(user_opts.play_pause_mbtn_left_command)`
3. Possible double cycle?

Since `_down` is being defined, is step `2.` still needed?

```diff
    ne.eventresponder["mbtn_left_up"] = function ()
        if state.eof_reached then
            mp.commandv("seek", 0, "absolute-percent")
            mp.commandv("set", "pause", "no")
-       else
-           mp.command(user_opts.play_pause_mbtn_left_command)
        end
    end
```

Keith94:
If I remove those lines (`else` and `mp.command`) it doesn't work properly. I don't see any conflict here unless I'm wrong.

Samillion:
But it does override, no? (conflict)

Let me show how to test what I mean, I'm bad at wording things.

In `modernz.conf`:

```EditorConfig
play_pause_mbtn_left_command=show-text "test"
```

Revert back the `mp.commandv()` but keep `bind_buttons()` where it is:
```lua
    bind_buttons("play_pause") -- dead code. always overridden by event below it
    ne.eventresponder["mbtn_left_up"] = function ()
        if state.eof_reached then
            mp.commandv("seek", 0, "absolute-percent")
            mp.commandv("set", "pause", "no")
        else
            mp.commandv("cycle", "pause")
        end
    end
```

See what I mean?

Samillion:
Hmmm. It works though. It's not actually deadcode.

It'll handle other clicks, and it's overridden where needed

```lua
bind_buttons("play_pause") -- mbtn_right and shift+mbtn_left (mid) and mbtn_left_up is overridden below
ne.eventresponder["mbtn_left_up"] = function ()
    if state.eof_reached then
        mp.commandv("seek", 0, "absolute-percent")
        mp.commandv("set", "pause", "no")
    else
        mp.command(user_opts.play_pause_mbtn_left_command) -- ensures no conflict
    end
end
```

Samillion:
Thank you very much.

Merging.

PR #691, opened 2026-04-19
feat: add URL decoding for media titles

Nagutos:
This PR fixes the issue where web streams without metadata display raw URL characters (like `%20` instead of spaces) in the OSC title.

- Added regex decoding in `make_escaped_title`.
- Wrapped in `if state.is_URL` to ensure local file paths remain completely unaffected.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -2873,6 +2873,13 @@ local function osc_init()
 
     local function make_escaped_title(source)
         local title = mp.command_native({"expand-text", source})
+        -- Decode URL title (%XX code)
+        if state.is_URL then
+            title = title:gsub("+", " ")
+            title = title:gsub("%%(%x%x)", function(h)
+                return string.char(tonumber(h, 16))
+            end)
+        end
         title = title:gsub("\n", " ")
         return title ~= "" and mp.command_native({"escape-ass", title}) or "mpv"
     end


Samillion:
Thank you for your effort.

- Could you provide an example URL that showcases this issue?
- Wouldn't this be more suitable as its own script that sets `--title` and/or `--script-opts-append=modernz-title`?
  - It would be a simple script: 
    - `on_load` hook
    - Check path
    - If it's url (and if title has encoded chars), decode
  - Good reference: https://gist.github.com/liukun/f9ce7d6d14fa45fe9b924a3eed5c3d99

Nagutos:
Hey,

To answer your question, here's the exact kind of URL causing the issue (it's generated by an debrid service extension and sent directly to mpv):
https://fake.debrid.it/dl/4m2tywqf6f0/movie%20example.mp4

Since it's a direct stream link, mpv has no metadata to read and just grabs the raw, encoded file path for the title.

But you're spot on—putting this inside ModernZ's core isn't the right move. A quick local on_load script that decodes the path and feeds it to ModernZ via modernz-title makes way more sense.

I'll go ahead and close the PR. Really appreciate the guidance!

Samillion:
If you make the script, let me know. I'll gladly link your repo in /extras and wherever relevant.

PR #692, opened 2026-04-19
pause-indicator: fix indicator showing on resize

Keith94:
The pause indicator was triggered every time window was resized. (follow up #683)

### DIFF
--- a/extras/pause-indicator-lite/pause_indicator_lite.lua
+++ b/extras/pause-indicator-lite/pause_indicator_lite.lua
@@ -69,6 +69,7 @@ local state = {
     toggled = false,
     eof = false,
     keybinds_registered = false,
+    indicator_dismissed = false,
 }
 
 local icon_theme = {
@@ -205,7 +206,7 @@ local function kill_timer(key)
 end
 
 local function update_indicator(force)
-    if state.aspect == 0 then return end
+    if not state.aspect or state.aspect == 0 then return end
     if not force and state.indicator_visible then
         return
     end
@@ -219,12 +220,13 @@ local function update_indicator(force)
         state.indicator_timer = mp.add_timeout(options.indicator_timeout, function()
             state.indicator_overlay:remove()
             state.indicator_visible = false
+            state.indicator_dismissed = true
         end)
     end
 end
 
 local function update_flash_icon()
-    if state.aspect == 0 or not options.flash_play_icon then return end
+    if not state.aspect or state.aspect == 0 or not options.flash_play_icon then return end
     kill_timer("flash_timer")
     state.flash_overlay:remove()
     state.flash_overlay.data = draw_triangle()
@@ -235,7 +237,7 @@ local function update_flash_icon()
 end
 
 local function update_mute_icon()
-    if state.aspect == 0 then return end
+    if not state.aspect or state.aspect == 0 then return end
     state.mute_overlay:remove()
     state.mute_overlay.data = draw_mute()
     state.mute_overlay:update()
@@ -277,6 +279,7 @@ local pause_observer = function(_, paused)
         kill_timer("indicator_timer")
         state.indicator_overlay:remove()
         state.indicator_visible = false
+        state.indicator_dismissed = false
         if state.toggled then
             update_flash_icon()
             state.toggled = false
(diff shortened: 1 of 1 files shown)


Samillion:
Thank you very much for the PR.

> The pause indicator was triggered every time window was resized.

Could you elaborate a bit, please?

You mean when not paused and resizing window (corners or maximize/unmaximize), the indicator shows for you?

Keith94:
Oops. Here's some steps to clarify.

1. Put `indicator_stay=no`
2. Pause a video
3. Let indicator disappear
4. Drag the corner to resize

Indicator pops up every time dimension changes

Samillion:
Ooo, nice catch. Would this be simpler? (Question, not trying to impose)

add to `local state`:

```lua
    dims_received = false,
```

Then:

```lua
local dimensions_observer = function()
    local _, _, aspect = mp.get_osd_size()
    state.aspect = aspect

    local dims_received = state.dims_received
    state.dims_received = true

    if state.paused and not state.indicator_visible and not dims_received then
        update_indicator(true)
        state.toggled = true
    elseif state.indicator_visible then
        update_indicator(true)
    end

    if state.mute_visible then
        update_mute_icon()
    end
end
```

Keith94:
> Ooo, nice catch. Would this be simpler? (Question, not trying to impose)

I tried, but it won't address the original issue I brought up: #683

Samillion:
Terribly sorry about the delay in response.

> I tried, but it won't address the original issue I brought up: https://github.com/Samillion/ModernZ/pull/683

Are you sure? It's working fine for me: (Could you elaborate more, please. I know you're right, I just can't reproduce)

image

Keith94:
I checked again, but my results are inconsistent. It shows the paused state correctly on file load maybe 50% of the time? I have no idea why. Checked using your method and this PR method gives me the same result.

Can you try double clicking a file to launch it from Explorer if that makes any difference?

https://github.com/user-attachments/assets/4aff629e-9880-4c6b-b379-0d76d17d452b

Samillion:
Wait, is this also with `indicator_stay=no`?

Keith94:
> Wait, is this also with `indicator_stay=no`?

Yes sir I had that set in my config.

Samillion:
Ok, this makes more sense now.

First: This PR issue and https://github.com/Samillion/ModernZ/pull/683 is fixed with the suggested edit, right?

Second:
The inconsistency with `indicator_stay=no` and `watch-later/pause` might be due to flash timer. As in, when you re-open the file, the icon flash already happened.

Genuine question: If so, would that be a bug? I need to think about this and also test a few possible scenarios

Keith94:
>First: This PR issue and https://github.com/Samillion/ModernZ/pull/683 is fixed with the suggested edit, right?

Showing on resized is fixed, but showing paused state on load works 50% of the time for me.
In https://github.com/Samillion/ModernZ/pull/692#issuecomment-4280549337 I showcase it.

Actually if I set `indicator_stay=yes` the outcome is the same.

Samillion:
The aspect check you made works consistently for you? (I'm unable to reproduce, so I have to rely on your feedback)

Sorry for the constant back and forth with checks, just trying to see if anything can be adjusted/improved.

Keith94:
>The aspect check you made works consistently for you?

Not really, it works inconsistently too. So the core issue just lies somewhere else.

Samillion:
Would resetting it in file loaded event do the trick?

Add to `local state` (end):

```lua
    dims_received = false,
```

Reset it in `file-loaded` event:

```diff
        state.eof = false
        state.paused = false
+       state.dims_received = false
```

`dimensions_observer`:

```lua
local dimensions_observer = function()
    local _, _, aspect = mp.get_osd_size()
    local first_dims = not state.dims_received
    state.dims_received = true
    state.aspect = aspect
    if aspect == 0 then return end

    if first_dims and state.paused and not state.indicator_visible then
        update_indicator(true)
        state.toggled = true
    elseif state.indicator_visible then
        update_indicator(true)
    end

    if state.mute_visible then
        update_mute_icon()
    end
end
```

Keith94:
> Would resetting it in file loaded event do the trick?

Sadly won't work, but seems [c16208d](https://github.com/Samillion/ModernZ/pull/692/commits/c16208d049ff17ed3fa7e1646c973eabec1bfb24) does work reliably for me. Maybe can be simplified.

Samillion:

Merging.

PR #700, opened 2026-04-22
feat: add seekbar wheel actions

### LINKED ISSUE #595: Seekbar scroll wheel amount

As a suggestion, since the scroll wheel behavior on seekbar is hardcoded to 10 seconds, should it allow customization?

    ne.eventresponder["wheel_up_press"] = function () mp.commandv("seek", 10) end
    ne.eventresponder["wheel_down_press"] = function () mp.commandv("seek", -10) end

For example, should it follow `jump_amount` and `jump_mode` etc. instead? Or have its own setting?

Keith94:
Fix #595


Samillion:
Thank you very much. Merging.

PR #701, opened 2026-04-22
feat: let cursor hide if keep_with_cursor disabled

### LINKED ISSUE #542: Allow cursor autohide with keep_with_cursor=no while hovering osc
### Reproduction Steps

1. `keep_with_cursor=no` in `modernz.conf`
2. `cursor-autohide=2000` in `mpv.conf`
3. Hover the pause button and let the osc fade out.

### Expected Behavior

The cursor autohides 2 seconds after mouse cursor stops

### Actual Behavior

The cursor autohides 2 seconds after osc fully fades out

### Log File

_No response_

Keith94:
Fixes #542

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -619,6 +619,7 @@ local state = {
     osd = mp.create_osd_overlay("ass-events"),
     logo_osd = mp.create_osd_overlay("ass-events"),
     keeponpause_active = false,             -- keeponpause bottombar active state
+    cursor_hide_flag = nil,                 -- flag to add allow-hide-cursor when keep_with_cursor=no
     chapter_list = {},                      -- sorted by time
     chapter = -1,                           -- current chapter index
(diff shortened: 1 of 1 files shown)


Samillion:
😆 This is the method I disliked that I mentioned here https://github.com/Samillion/ModernZ/issues/542#issuecomment-4181851912

It works, but it's spammy, for lack of better words. That's why I mentioned here https://github.com/Samillion/ModernZ/issues/542#issuecomment-4284884928 why I didn't implement it.

Keith94:
Ahhh okay. I misunderstood you lol. Nevermind

Samillion:
There are two points fighting in my head:
- While I say I dislike it, I still can't figure out an alternative method
- Even if it works, it's still spammy. By that I mean, multiple clicks (ie: jump button), will generate 4x calls per click, and that's just the part I'm aware of

Keep it as a draft, please. We might figure out a better implementation.

PR #740, opened 2026-05-01
refactor: remove unused maxchars dead code

Samillion:
**Changes**:
- Remove unused `maxchars` dead code
  - All this time `maxchars` was set to `nil`, so the condition is never true and never runs
  - ModernZ relies on clipping the title(s) based on window width, not maximum characters count
  - This works in `osc.lua` with fixed width layouts, ours is more dynamic and adjusts based on window size

**Note**:
While we can simply assign an arbitrary value to `maxchars` (e.g. `150` characters), the change it applies looks unpleasant. It's not just a simple adding of three dots at the end. It uses `\\fscx` to shrink it, which makes the title look compressed and weird.

However, I'm open to suggestions, this doesn't have to be removed if it can be used properly.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -1771,19 +1771,6 @@ local function render_elements(master_ass, osc_vis, wc_vis)
                 buttontext = element.content -- text objects
             end
 
-            local maxchars = element.layout.button.maxchars
-            if maxchars ~= nil and #buttontext > maxchars then
-                local max_ratio = 1.25  -- up to 25% more chars while shrinking
-                local limit = math.max(0, math.floor(maxchars * max_ratio) - 3)
-                if #buttontext > limit then
-                    while (#buttontext > limit) do
-                        buttontext = buttontext:gsub(".[\128-\191]*$", "")
-                    end
-                    buttontext = buttontext .. "..."
-                end
-                buttontext = string.format("{\\fscx%f}", (maxchars/#buttontext)*100) .. buttontext
-            end
-
             -- add hover effects
             local button_lo = element.layout.button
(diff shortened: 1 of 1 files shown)


Samillion:
Alternatively, should we turn this `maxchars = nil,` into a user option to be controlled, `nil` by default?

It was requested here https://github.com/Samillion/ModernZ/discussions/726 by @yannikthegenius (it's how I realized this was deadcode all this time)

My objection isn't the feature itself, it's how it makes the title(s) look.

Obligatory cc to @Keith94 and @nekoxuee

Keith94:
>It uses \\fscx to shrink it, which makes the title look compressed and weird.

The shrinking is probably not the best idea. It's better to have only ellipsis when the title exceeds. I was playing around with it here: https://github.com/Keith94/ModernZ/commit/b1b8dc42b01f2a81b468a49df4308a33aa2590ee

User option is likely not necessary.

nekoxuee:
I agree that shrinking is undesirable.

Samillion:
> The shrinking is probably not the best idea. It's better to have only ellipsis when the title exceeds. I was playing around with it here: [Keith94@b1b8dc4](https://github.com/Keith94/ModernZ/commit/b1b8dc42b01f2a81b468a49df4308a33aa2590ee)

Beautiful, much cleaner.

~~Question: should the states be initiated as an infinite number? In other words, can they be initiated as nil or 0, then assigned value on actual use/render instead?~~

Either way, I'll merge this, then you feat PR your idea, sounds good?

Keith94:
> Either way, I'll merge this, then you feat PR your idea, sounds good?

Okay, sounds great. I'll do it tomorrow.

PR #742, opened 2026-05-02
feat: add ellipsis to long title texts

Keith94:
Reference: https://github.com/Samillion/ModernZ/discussions/726

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -2493,7 +2493,7 @@ layouts["compact"] = function ()
     end
 
     -- osc title
-    local title_w = (no_chapter or not chapter_index or user_opts.chapter_above_title) and (osc_geo.w - 60 - time_codes_width) or (osc_geo.w - 50)
+    local title_w = osc_geo.w - 50
     state.title_max_w = title_w
     if title_w < 0 then title_w = 0 end
     elements["title"].visible = not no_title
--- a/modernz.lua
+++ b/modernz.lua
@@ -647,6 +647,9 @@ local state = {
     downloaded_once = false,
     downloading = false,
     file_size_normalized = nil,
+    title_max_w = nil,
+    windowtitle_max_w = nil,
+    chapter_title_max_w = nil,
 }
 
 local logo_lines = {
@@ -2069,6 +2072,7 @@ local function window_controls()
         lo.group = "top"
         lo.alpha[3] = 0
         lo.style = string.format("%s{\\clip(%f,%f,%f,%f)}", osc_styles.window_title, titlebox_left, wc_geo.y - wc_geo.h, controlbox_left, wc_geo.y + wc_geo.h)
+        state.windowtitle_max_w = controlbox_left - math.max(20, titlebox_left + 4)
     end
 
     -- only add top areas and margin if one of the elements is enabled
@@ -2219,6 +2223,7 @@ layouts["modern"] = function ()
     -- osc title
     local title_w = (no_chapter or not chapter_index or user_opts.chapter_above_title) and (osc_geo.w - 60 - time_codes_width) or (osc_geo.w - 50)
     if title_w < 0 then title_w = 0 end
+    state.title_max_w = title_w
     elements["title"].visible = not no_title
     geo = {x = 25, y = refY - title_y, an = 1, w = title_w, h = user_opts.title_font_size}
     lo = add_layout("title")
@@ -2236,6 +2241,7 @@ layouts["modern"] = function ()
         lo.layer = 48
         lo.alpha[3] = 0
         lo.style = string.format("%s{\\clip(%f,%f,%f,%f)}", osc_styles.chapter_title, 0, 0, geo.x + geo.w, geo.y + geo.h)
+        state.chapter_title_max_w = geo.w
     end
 
     -- buttons
@@ -2478,6 +2484,7 @@ layouts["modern-compact"] = function ()
     -- osc title
     local title_w = (no_chapter or not chapter_index or user_opts.chapter_above_title) and (osc_geo.w - 60 - time_codes_width) or (osc_geo.w - 50)
     if title_w < 0 then title_w = 0 end
+    state.title_max_w = title_w
     elements["title"].visible = not no_title
     geo = {x = 25, y = refY - title_y, an = 1, w = title_w, h = user_opts.title_font_size}
     lo = add_layout("title")
@@ -2495,6 +2502,7 @@ layouts["modern-compact"] = function ()
         lo.layer = 48
         lo.alpha[3] = 0
         lo.style = string.format("%s{\\clip(%f,%f,%f,%f)}", osc_styles.chapter_title, 0, 0, geo.x + geo.w, geo.y + geo.h)
+        state.chapter_title_max_w = geo.w
     end
 
     -- time codes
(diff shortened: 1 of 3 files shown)


Samillion:
So much better than the shrinking behavior `maxchars` had, beautifully done.

And you even indulged my annoying state initiation comment. ❤️

Keith94:
Thanks, and yeah I like the result. I kinda messed up the commits and am trying to squash it.

Samillion:
> and am trying to squash it.

Doesn't matter, in the end all is well, it's just semantics. I even deleted my comment about it before. It's nice that we have a form to follow, but mistakes happen.

Keith94:
All right sorry about that. It won't let me fix unless I open a new one. :(

Samillion:
You're way overthinking this, no worries at all. The work is solid and well done, that's all that matters. ❤️

Samillion:
Questions: (Because I'm genuinely not sure)
Don't make any changes, I'm genuinely not sure if any of this is actually of concern or just me overthinking stuff as usual.

## Question `#1`

Does the calculation with ellipsis has to be iterated on each calculation? Since it's a preset, we can do something like:

```lua
-- add
local ell_w = estimate_text_width("…", style)
```

Then:

```lua
-- replace
if estimate_text_width(title:sub(1, mid), style) <= max_w - ell_w then
```

But this is trivial, right? No actual performance impact or gain here.... I think.

## Question `#2`

Since this calculates bytes, how would it handle characters like in Simplified Chinese or Japanese? If I remember correctly, a Chinese character is 3 bytes? The truncated string could have a malformed sequence at the cut point, If `fit` lands mid point... I think... right? 

TLDR: Should `fit` go back to the nearest valid UTF-8 boundary after the search? Or is this a waste and unnecessary addition? 

Adding something like: 

```lua
        while fit > 0 and (title:byte(fit) or 0) >= 0x80 and (title:byte(fit) or 0) < 0xC0 do
            fit = fit - 1
        end
```

## Full picture

In essence, this is the full helper with what I'm talking about:

```lua
    local function truncate_title(title, max_w, style)
        if not max_w or max_w <= 0 or estimate_text_width(title, style) <= max_w then return title end
        local ell_w = estimate_text_width("…", style)
        local low, high, fit = 1, #title, 0
        while low <= high do
            local mid = math.floor((low + high) / 2)
            if estimate_text_width(title:sub(1, mid), style) <= max_w - ell_w then
                fit = mid; low = mid + 1
            else
                high = mid - 1
            end
        end
        while fit > 0 and (title:byte(fit) or 0) >= 0x80 and (title:byte(fit) or 0) < 0xC0 do
            fit = fit - 1
        end
        return title:sub(1, fit) .. "…"
    end
```

Samillion:
cc @nekoxuee terribly sorry for bothering you, would love your input on this as well.

Keith94:
>Question 1

Calculating it once would be a good optimization.

>Question 2

You're right that it would look malformed. I'll just test if that fixes it.

Keith94:
This addresses CJK character issue.

``` diff
- local low, high, fit = 1, #title, 0
+ local ell_w = estimate_text_width("…", style)
+ -- map each UTF-8 character to its last byte offset (avoids slicing mid-char)
+ local char_ends, pos = {}, 1
+ while pos <= #title do
+     local b = title:byte(pos)
+     local char_len = b >= 0xF0 and 4 or b >= 0xE0 and 3 or b >= 0xC0 and 2 or 1
+     char_ends[#char_ends + 1] = pos + char_len - 1
+     pos = pos + char_len
+ end
+ -- binary search over character count
+ local low, high, fit = 1, #char_ends, 0

- if estimate_text_width(title:sub(1, mid) .. "…", style) <= max_w then
+ if estimate_text_width(title:sub(1, char_ends[mid]), style) <= max_w - ell_w then

- return title:sub(1, fit) .. "…"
+ return title:sub(1, fit > 0 and char_ends[fit] or 0) .. "…"
```

nekoxuee:
> ## Question `#1`
> 
> Does the calculation with ellipsis has to be iterated on each calculation? Since it's a preset, we can do something like:
> 
> ```lua
> -- add
> local ell_w = estimate_text_width("…", style)
> ```
> 
> Then:
> 
> ```lua
> -- replace
> if estimate_text_width(title:sub(1, mid), style) <= max_w - ell_w then
> ```
> 
> But this is trivial, right? No actual performance impact or gain here.... I think.

shouldn't have a meaningful impact since estimate_text_width() has cache but still correct doing it.

> Since this calculates bytes, how would it handle characters like in Simplified Chinese or Japanese? If I remember correctly, a Chinese character is 3 bytes? The truncated string could have a malformed sequence at the cut point, If `fit` lands mid point... I think... right?
> 
> TLDR: Should `fit` go back to the nearest valid UTF-8 boundary after the search? Or is this a waste and unnecessary addition?

yes, this is a real bug, should be fixed.

modernz.lua
```diff
@@ -2486,6 +2492,7 @@ layouts["compact"] = function ()
     end
 
     -- osc title
+    state.title_max_w = title_w
```
nekoxuee, line 2498: `title_w` is only defined as a local variable inside the default layout

Keith94: Fixed. Thanks!

Keith94: Perhaps I misunderstood but the compact layout still needs this, no?

nekoxuee: It does, what I said is that title_w is defined inside the default layout function, so its out of scope for the compact layout. Because of that title_w is nil inside the compact layout and truncate_title never runs.

Samillion:
> This addresses CJK character issue.

Does my suggestion not work? It's much simpler. (Asking, not trying to impose which one would work).

Either way, whichever you deep fit, go for it.

Samillion:
Sorry for the additional comment in a row, could you please tie this feature to an option?

`truncate_title=yes/no`

Disabled by default.

Keith94:
>Does my suggestion not work?

Tested it, sadly did not work.

>could you please tie this feature to an option?

Added `truncate_title = false`

Samillion:
You are awesome, thank you very much. ❤️ 

Let me know when it's merge ready, please. No rush at all.

Keith94:
Added missing `\\q2` to chapter title so it won't wrap to a new line. <s>Should be good to merge.</s>

Keith94:
Fixed a couple width issues. Hopefully it's good now.

Samillion:
> Added missing `\\q2` to chapter title so it won't wrap to a new line

What's the edge case that required this?  And only for chapter title?

https://aegisub.org/docs/latest/ass_tags/#\q

Keith94:
> > Added missing `\\q2` to chapter title so it won't wrap to a new line
> 
> What's the edge case that required this? And only for chapter title?
> 
> [https://aegisub.org/docs/latest/ass_tags/#\q](https://aegisub.org/docs/latest/ass_tags/#%5Cq)

If you have a long chapter title it could overlap with osc title in some situations. If you want to test:

```editorconfig
vidscale=no
chapter_title_font_size=30
title_font_size=36
# mpv.conf
geometry=650x375
```

Video: https://youtu.be/uM-fS9bRiAo?t=1576

mpv_5gjvET9uEJ

Title and window title have this already set in `osc_styles`.

```editorconfig
        window_title = "{\\bord1\\1c&H" .. osc_color_convert(user_opts.window_title_color) .. "&\\3c&H0&\\fs".. user_opts.window_title_font_size .."\\q2\\fn" .. user_opts.font .. "}",
        title = "{\\bord1\\1c&H" .. osc_color_convert(user_opts.title_color) .. "&\\3c&H0&\\fs".. user_opts.title_font_size .."\\q2\\fn" .. user_opts.font .. "}",
```

Samillion:
Terribly sorry about the delay, currently dealing with urgent matters.

Merging.

PR #744, opened 2026-05-02
feat: add generic wheel scroll and press commands

Keith94:
Commands set as default actions are applied to elements that have no explicit command defined.

Reference: https://github.com/Samillion/ModernZ/issues/698

PR #747, opened 2026-05-08
refactor: split has_click_action for mouse actions

Keith94:
Wheel up, down, and mbtn_mid get their own areas so unhandled events fall through to input.conf.

Alternative to #744 which adds no user options but I'm not sure which method is preferred.

### DIFF
--- a/modernz.lua
+++ b/modernz.lua
@@ -3731,7 +3731,6 @@ end
 local click_keys = {
     "mbtn_left_up", "mbtn_left_down", "mbtn_left_press",
     "mbtn_right_up", "mbtn_right_down", "mbtn_right_press",
-    "wheel_up_press", "wheel_down_press",
 }
 local function has_click_action(e)
     if not e.eventresponder then return false end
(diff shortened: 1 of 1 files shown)


Samillion:
I apologize again for the delay.

If I may ask and be a burden, have you tested this thoroughly?

I don't want to delay further, as I've inconvenienced your efforts enough already. Do you need me to help with anything in particular?

I'll be back to this again with full focus soon. Just need some time. I hope you understand.

Keith94:
Oh hai. I tested a lil bit and didn't see any bugs. After more tests I'll let you know.

Samillion:
Thank you very much. Take all the time you need.

Keith94:
Everything works as expected. Wheel up/down won't fall through to input.conf unless both up and down commands are disabled for that button. Same as how left and right click works.

Samillion:
Thank you very much. I've tested all scenarios I can think of and everything works very well.

Merging.

PR #755, opened 2026-05-22
Add title resolution for YouTube videos

edieguez:
- Resolve URLs titles using `yt-dlp`
- This behaviour can be disabled by changing `resolve_url_titles` to `no`

### DIFF
--- a/modernz.conf
+++ b/modernz.conf
@@ -173,6 +173,8 @@ screenshot_button=yes
 download_button=yes
 # default download directory for videos (https://mpv.io/manual/master/#paths)
 download_path=~~desktop/mpv
+# resolve titles for URL playlist entries via yt-dlp
+resolve_url_titles=yes
 
 # show file loop button
 loop_button=yes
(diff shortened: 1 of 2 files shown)


Samillion:
Thanks for the contribution, but this falls outside the scope of what ModernZ should handle.

mpv already resolves playlist titles natively. From the manual:
https://github.com/mpv-player/mpv/blob/master/DOCS/man/input.rst

- `media-title` uses the file's title tag, falling back to the filename
- `playlist/N/title` is populated automatically once an entry has been played, or if the playlist format includes it

Your implementation actually reads `playlist/N/title` first, which confirms that mpv's own mechanism works.

The yt-dlp fallback is only filling the gap for entries not yet played. Pre-fetching titles for unplayed playlist entries is not something an OSC should be responsible for, especially when it means spawning one `yt-dlp` subprocess per untitled entry on every `playlist-count` change.

If this is something you'd like to pursue, it would be better suited as a standalone script that handles playlist pre-fetching independently of the OSC.

